PR #7, opened 2025-06-20
Add Pydantic output models for integrals (S, H, F,  J, and K)

haneug:
Also updates the Fock matrix diagonalization notebook to use the output instead of directly loading the json.

### DIFF
--- a/src/opi/output/models/json/gbw/properties/molecule.py
+++ b/src/opi/output/models/json/gbw/properties/molecule.py
@@ -1,4 +1,4 @@
-from pydantic import Field, StrictInt, StrictStr
+from pydantic import Field, StrictFloat, StrictInt, StrictStr
 
 from opi.output.models.base.get_item import GetItem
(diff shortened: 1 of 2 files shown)


src/opi/output/models/json/gbw/properties/molecule.py
```diff
@@ -21,9 +21,23 @@ class Molecule(GetItem):
     molecularorbital: MolecularOrbital
         Contains information about the molecular orbitals
     multiplicity: StrictInt
-        multiplicity of the atom
+        multiplicity of the molecule
+    charge: StrictInt
+        charge of the molecule
     hftyp: StrictSrt
         used type of HF in the calculation
+    origin: tuple[StrictFloat,StrictFloat,StrictFloat]
```
timmyte: Please add spaces between the inner types of the tuple. Like here: `origin: tuple[StrictFloat, StrictFloat, StrictFloat]`

PR #10, opened 2025-06-23
Add function for accessing optimized structures more easily from the Output class

haneug:
- Adds the `get_structure` function to Output to obtain a structure based on an index.
- Adds a `_get_cartesians` function to verify that Cartesian coordinates are available. 

The tree up to Cartesians has to be checked instance by instance to make mypy happy. I think this will be required for any easy access functions. Therefore, we should probably add something like a `safe_get(obj: Any, *attrs: str) -> Any `  function for easier access to instances within the output class tree.

### DIFF
--- a/src/opi/output/core.py
+++ b/src/opi/output/core.py
@@ -20,7 +20,7 @@ from opi.output.models.json.property.property_results import (
 )
 from opi.utils.misc import check_minimal_version, lowercase
 from opi.utils.orca_version import OrcaVersion
-from opi.utils.units import au_to_angst
+from opi.utils.units import AU_TO_ANGST
 
 
 class Output:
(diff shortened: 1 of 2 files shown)


src/opi/utils/units.py
```diff
@@ -0,0 +1,4 @@
+au_to_angst: float = 0.52917721067
```
timmyte: Constants should be in all-caps.

src/opi/output/core.py
```diff
@@ -314,3 +319,94 @@ def terminated_normally(self) -> bool:
...
+    def _safe_get(self, *attrs: str | int) -> Any | None:
+        """
+        Safely access a nested chain of attributes or list indices to access results from the output object.
+
+        This function walks through a sequence of attributes (str) or indices (int),
+        returning the nested value if all lookups succeed and none of the intermediate
+        objects are None. If any step fails (due to missing attribute, invalid index,
+        or None in the chain), it returns None instead of raising an exception.
+
+        Parameters:
```
timmyte: Please use numpydocstyle for the docstring.

src/opi/output/core.py
```diff
@@ -314,3 +319,94 @@ def terminated_normally(self) -> bool:
...
+                    raise TypeError
+            except (AttributeError, IndexError, TypeError):
+                return None
+        return current
+
+    def _get_cartesians(
+        self, index: int
+    ) -> list[tuple[StrictStr, StrictFiniteFloat, StrictFiniteFloat, StrictFiniteFloat]] | None:
+        """
+        Returns cartesian coordinates from the output object for a specified index
```
timmyte: Please add the `Parameter`, `Returns` and `Raises` sections to the docstring.

Issue #11, opened 2025-06-24
Check for succesful geometry optimization


haneug:
**Is your feature request related to a problem? Please describe.**
Currently, opi can check for the normal termination of ORCA and the single_point_data group in the output contains information whether the SCF has converged. A simple function for checking if a geometry optimization has converged would be a useful addition.

**Describe the solution you'd like**
The recipe is already there. Just add a geometry_optimization_converged() function to the output, while at this one can also add a convenience function for the SCF.

PR #12, opened 2025-06-24
Add check function for succesful geometry optimization to output

haneug:
Related issue #11

### DIFF
--- a/examples/exmp003_opt/job.py
+++ b/examples/exmp003_opt/job.py
@@ -2,6 +2,7 @@
 
 import shutil
 from pathlib import Path
+import sys
 
 from opi.core import Calculator
 from opi.input.simple_keywords import BasisSet, Dft, Scf, Task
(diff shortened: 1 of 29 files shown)


examples/exmp003_opt/job.py
```diff
@@ -29,6 +29,16 @@
     # > Parse JSON files
     output.parse()
 
+    # > Verify that SCF converged
+    if not output.scf_converged():
+        print(f"ORCA SCF failed to converge, see output file: {output.get_outfile()}")
+        exit(1)
```
timmyte: Preferred way to exit in scripts is to use `sys.exit()`.

examples/exmp003_opt/job.py
```diff
@@ -29,6 +29,16 @@
...
+
+    # > Verify that geometry optimization converged
+    if not output.geometry_optimization_converged():
+        print(f"ORCA geometry optimization failed to converge, see output file: {output.get_outfile()}")
+        exit(1)
```
timmyte: See comment above.

haneug:
Good point, I updated the exit statements in the examples to sys.exit()

PR #32, opened 2025-06-26
Add functions for easier access to energies in output

### LINKED ISSUE #20: Function for easier access to energies
In the output there is a list containing different energies such as SCF, CASSCF, or different Post-SCF energies. It would be nice to have a convenience function that gives you a better overview of these energies.

Put the energies into a dict with the energy type as key and the objects as values.

haneug:
- get_final_energy(), for simple access to final energy
- get_energies(), for access to different energy contributions. The list of energies is converted into a dictionary where the key indicates the energy type. If multiple energies of the same type occur, an index is added at the end of the key (not sure where this would occur).

closes #20


src/opi/output/core.py
```diff
@@ -478,3 +480,81 @@ def get_structure(self, *, index: int = -1) -> Structure:
...
+        Easy access to the final single point energy
+
+        Returns
+        ----------
+        final_energy: StrictFiniteFloat
+            Returns the final energy of the ORCA calculation
+
+        Raises
+        ----------
+        ValueError
```
timmyte: I would just return `None`.

src/opi/output/core.py
```diff
@@ -478,3 +480,81 @@ def get_structure(self, *, index: int = -1) -> Structure:
...
+
+        Returns
+        -------
+        energy_dict : dict[str, Energy]
+            Dictionary where keys identify the energy type. If multiple energies of the same type are present, an index is
+            appended after the first one, e.g., SCF, SCF_1, SCF_2, etc.
+
+        Raises
+        ----------
+        ValueError
```
timmyte: I would just return `None`.

src/opi/output/core.py
```diff
@@ -489,3 +491,69 @@ def get_structure(self, *, index: int = -1) -> Structure:
...
+        Returns
+        ----------
+        final_energy: StrictFiniteFloat | None
+            Returns the final energy of the ORCA calculation or None if there is none in the output.
+        """
+
+        # > Get the final energy
+        final_energy = self._safe_get("results_properties", "single_point_data", "finalenergy")
+
+        if final_energy:
```
timmyte: Can the energy also be `0.0`? Also consider non-trivial cases.
It would be safer to check for `if final_energy is not None` here.

src/opi/output/core.py
```diff
@@ -489,3 +491,69 @@ def get_structure(self, *, index: int = -1) -> Structure:
...
+
+        Returns
+        -------
+        energy_dict : dict[str, Energy] | None
+            Dictionary where keys identify the energy type. If multiple energies of the same type are present, an index is
+            appended after the first one, e.g., SCF, SCF_1, SCF_2, etc. If no energy is available None is returned.
+
+        Notes
+        -----
+        Common keys include: **Unknown** - No information about the energy is provided,
```
timmyte: One key per line please:
```
Common keys include: 
    - **Unknown** ...
    - **SCF** ....
    - ...
```

src/opi/output/core.py
```diff
@@ -489,3 +491,69 @@ def get_structure(self, *, index: int = -1) -> Structure:
...
+        # > Energy dict to populate & return
+        energy_dict: dict[str, Energy] = {}
+
+        # > Get the list of energies to be converted to a dictionary
+        energy_list = self._safe_get("results_properties", "geometries", index, "energy")
+
+        if energy_list:
+            energy_list = cast(EnergyList, energy_list)
+        else:
+            return energy_list
```
timmyte: `return None`. Be explicit.

timmyte (changes requested):

src/opi/output/core.py
haneug: Very good point!

src/opi/output/core.py
haneug: I also added empty lines in between, so that these are separate lines in the AutoAPI.

src/opi/output/core.py
```diff
@@ -489,3 +491,80 @@ def get_structure(self, *, index: int = -1) -> Structure:
...
+
+        Returns
+        -------
+        energy_dict : dict[str, Energy] | None
+            Dictionary where keys identify the energy type. If multiple energies of the same type are present, an index is
+            appended after the first one, e.g., SCF, SCF_1, SCF_2, etc. If no energy is available None is returned.
+
+        Notes
+        -----
+        Common keys include:
```
timmyte, line 530: Please arrange enumeration as follows:
```
Common keys include:
   - **Unknown** : No information about the energy is provided.
   - **SCF** : SCF energy from HF, DFT, or SQM methods.
   - **MDCI(SD)** : Typically the (DLPNO-)CCSD energy.
   - **MDCI(SD(T))** : Typically the (DLPNO-)CCSD(T) energy.
   - **CASSCF** : CASSCF energy.
   - **MP2** : MP2 energy.
   - **TDA/CIS** : TDA-TD-DFT or CIS energy.
```

PR #33, opened 2025-06-26
Switch for reading certain JSON files

### LINKED ISSUE #18: Selective reading of JSON files
Currently, OPI attempts to read various JSON files by default. However, this can cause issues, e.g., if no .gbw file is produced, OPI crashes when trying to read the corresponding JSON-gbw file.
It would be nice to have an option to specify which JSON files should be read (or read not) during processing of the output.

cplett:
## Feature
Introducing two optional switches to the `parse()` function to switch off the reading of the JSON-property and JSON-gbw file. This becomes important when no gbw files are created by ORCA, e.g., when using a force field or external methods via wrapper script. The switches are optional and per default `True`. 

## Usage
When using the `parse()` function of the output object, set the respective variable `False`. Variable names are:
- `read_prop_json` for JSON-property file
- `read_gbw_json` for JSON-gbw file

## Additional changes
As not reading the JSON-gbw file causes the `numofbasisfuncts` to be 0, I changed the `StrictPositiveInt` check to `StrictNonNegativeInt`.

Closes #18

### DIFF
--- a/src/opi/output/core.py
+++ b/src/opi/output/core.py
@@ -122,12 +122,19 @@ class Output:
         if parse:
             self.parse()
 
-    def parse(self) -> None:
+    def parse(self, read_prop_json: bool = True, read_gbw_json: bool = True) -> None:
         """
         Create property- and gbw-JSON file (according to `do_create_property_json` and `self.do_create_gbw_json`)
         and parse them.
-        """
+        Skips the parsing for the gbw or prop json file when the respective bool is false. It defaults to true
 
+        Parameters
+        ----------
+        read_prop_json: bool
+            Whether or not to read the property JSON file
+        read_gbw_json: bool
+            Whether or not to read the gbw JSON file
+        """
         # // Create JSONs files
         if self.do_create_gbw_json:
             self.create_gbw_json()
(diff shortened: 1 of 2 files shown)


src/opi/output/core.py
```diff
@@ -113,30 +113,38 @@ def __init__(
...
+    def parse(self, read_prop_json: bool = True, read_gbw_json: bool = True) -> None:
         """
         Create property- and gbw-JSON file (according to `do_create_property_json` and `self.do_create_gbw_json`)
         and parse them.
-        """
+        Skips the parsing for the gbw or prop json file when the respective bool is false. It defaults to true
 
+        Parameters
+        ----------
+        read_prop_json: bool
```
timmyte: State default value here to mark them as optional:
```
read_prop_json: bool, default: True
```

src/opi/output/core.py
```diff
@@ -113,30 +113,38 @@ def __init__(
...
-        self.gbw_json_data = self._process_json_file(self.gbw_json_file)
-        self.property_json_data = self._process_json_file(self.property_json_file)
-
-        # > `property_json_data` needs to be set
-        if self.do_version_check:
-            self.check_version()
-
         # // Property JSON
-        self.results_properties = PropertyResults(**self.property_json_data)
+        if read_prop_json:
```
timmyte, line 146: I would probably print warning `warn()` if  `read_prop_json` is `False`, but `do_version_check` is `True`.
Something along the line: "Cannot perform version check".
Decision is up to you, but this should definitely be mentioned in the docstring that `read_prop_json` overrules `do_version_check`.

timmyte (changes requested):

src/opi/output/core.py
cplett: Done

src/opi/output/core.py
cplett: Good point. Warning is included

PR #34, opened 2025-06-26
Add Fragment IDs to get_structure output function

### LINKED ISSUE #23: Get structure with fragment IDs
Currently, the get_structure function in the output only returns the Cartesian coordinates. It should also check whether fragment IDs are present in the output and add them to the created structure object.

Add check whether fragment information is available and add it. Maybe, an optional bool for ignoring the fragment data could be useful.

haneug:
Adds fragment IDs to get_structure convenience function in the output. Can be turned off with optional bool `with_fragments`

closes #23
depends on #32

### DIFF
--- a/src/opi/output/core.py
+++ b/src/opi/output/core.py
@@ -17,7 +17,7 @@ from opi.output.grepper.recipes import (
     has_scf_converged,
     has_terminated_normally,
 )
-from opi.output.models.base.strict_types import StrictFiniteFloat
+from opi.output.models.base.strict_types import StrictFiniteFloat, StrictPositiveInt
 from opi.output.models.json.gbw.gbw_results import GbwResults
 from opi.output.models.json.property.properties.energy import Energy
 from opi.output.models.json.property.properties.energy_list import EnergyList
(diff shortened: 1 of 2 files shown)


src/opi/output/core.py
```diff
@@ -435,14 +437,43 @@ def _get_cartesians(
...
+        # > Cast them into the correct type
+        if fragments is not None:
+            fragments = cast(
+                list[list[StrictPositiveInt]],
+                fragments,
+            )
+
+        return fragments
+
+    def get_structure(self, *, index: int = -1, with_fragments = True) -> Structure:
```
timmyte: `with_fragments = True` type annotation missing.

src/opi/output/core.py
```diff
@@ -435,14 +437,43 @@ def _get_cartesians(
 
         return cartesians
 
-    def get_structure(self, *, index: int = -1) -> Structure:
+    def _get_fragments(self, index: int) -> list[list[StrictPositiveInt]] | None:
```
timmyte: Just a suggestion, I would make the `index` parameter positional-only, i.e. `_get_fragments(self, index: int, /)`.

src/opi/output/core.py
```diff
@@ -459,22 +490,106 @@ def get_structure(self, *, index: int = -1) -> Structure:
...
+        Easy access to the final single point energy
+
+        Returns
+        ----------
+        final_energy: StrictFiniteFloat
+            Returns the final energy of the ORCA calculation
 
+        Raises
+        ----------
+        ValueError
```
timmyte: I would just return `None`. This makes this getter more applicable.

src/opi/output/core.py
```diff
@@ -459,22 +490,106 @@ def get_structure(self, *, index: int = -1) -> Structure:
...
+
+        Returns
+        -------
+        energy_dict : dict[str, Energy]
+            Dictionary where keys identify the energy type. If multiple energies of the same type are present, an index is
+            appended after the first one, e.g., SCF, SCF_1, SCF_2, etc.
+
+        Raises
+        ----------
+        ValueError
```
timmyte: I would just return `None`. This makes this getter more applicable.

PR #35, opened 2025-06-27
Include Notebook for using OPI with wrapper scripts

cplett:


haneug (approved):
I am not completely sure about the usage of :::{note} ::: since it is not supported by default markdown viewers. However, it looks nice on the html pages.

PR #43, opened 2025-07-02
fix to fragproc attribute and additional functionality for basis block

nakul680:
- Allows multiple options for fragproc attribute in BlockFrag 
- Can now initialize FragBasis and associated attributes with string like other Block Inputs


src/opi/input/blocks/block_frag.py
```diff
@@ -66,8 +64,34 @@ class BlockFrag(Block):
             "delsolvents",
             "delwater",
         ]
-        | None
-    ) = None
+    ]
+
+    def __str__(self) -> str:
+        return " ".join(self.flags)
```
haneug: Keywords in FragProc are separated by a comma.

haneug (changes requested):
Note that Fragproc requires commas between multiple keywords. 
Could you create a minimal example for fragbasis? 
In the ORCA manual the following example can be found:
```
! PrintBasis BP86 NoIter
! def2-SVP def2/J
%basis
  FragBasis 1 "def2-TZVP"
  FragBasis 2 "cc-pVTZ-PP"
  FragAuxJ  2 "AutoAux"
  FragECP   3 "SK-MCDHF-RSC"
  FragAuxJ  3 "def2/JK"
end
*xyz 0 1
  H(1) 0 0 0
  I(1) 0 0 1.6
  H(2) 0 5 0  
  I(2) 0 5 1.6
  H(3) 5 0 0
  I(3) 5 0 1.6
*

haneug (changes requested):
Thanks for the nice example. Works nicely ~~but now when I import BlockFrag I get circular import issues: `ImportError: cannot import name 'InputFilePath' from partially initialized module 'opi.input.blocks' (most likely due to a circular import) `~~ Never mind, that was just my fault. Please run `uv run nox` at the end so everything is formatted as it should.

haneug (approved):
Please do `uv run nox` once again, otherwise this looks good!

PR #49, opened 2025-07-07
Fix for get_final_energy()

haneug:
- The path to the result in the get_final_energy function did not consider the geometries and would always return None.
- With this commit the geometries are considered in the path and optional int `index` is added.


src/opi/output/core.py
```diff
@@ -527,18 +527,25 @@ def get_structure(self, *, index: int = -1, with_fragments: bool = True) -> Stru
...
         return structure
 
-    def get_final_energy(self) -> StrictFiniteFloat | None:
+    def get_final_energy(self, *, index: int = -1) -> StrictFiniteFloat | None:
         """
         Easy access to the final single point energy.
 
+        Parameters
+        ----------
+        index : int, default: -1
```
timmyte, line 536: What happens if the given `index` is out of range?
I would describe this somehow in the docstring.

PR #50, opened 2025-07-07
Bugfix: %moinp block

### LINKED ISSUE #47: moinp is not properly working due to missing quotation marks
**Describe the bug**
When OPI writes %moinp to the ORCA input file, the quotation marks are missing and ORCA does not recognize the filename.

**To Reproduce**
Read a gbw file with another basename as initial guess via `calc.input.moinp = /Path/to/other/file.gbw` and run the calculation. Due to the missing quotation marks the ORCA calculation will fail. 

**Expected behavior**
Quotation marks should be set so that moinp works properly.

**Version:**
 - OS: [e.g. Ubuntu 24.04]
 - ORCA [e.g. 6.1.0]

haneug:
1. The %moinp block requires double quotation marks around the filename of the gbw file to read it.
2. The file can lie in the working directory (recommended) or somewhere else: If it lies within the working directory we write the relative Path (or just the filename) into the input file, and if it does not lie within the working directory we use the absolute path to access it.

closes #47


haneug:
Now by default relative paths are employed and a `ValueError` is raised if the relative path cannot be resolved or the file does not exist.

src/opi/core.py
```diff
@@ -221,7 +214,7 @@ def write_input(self, *, copy_structure: bool = True) -> None:
                 if (ncores := input_param.ncores) is not None:
                     inp.write(f"%pal\n    nprocs {ncores:d}\nend\n")
                 if (moinp := input_param.moinp) is not None:
-                    inp.write(f"%moinp {moinp}\n")
+                    inp.write(f'%moinp "{check_file_path(moinp, self.working_dir)}"\n')
```
timmyte: The name of `check_file_path()` is not reflecting the purpose of the function.
Pick something like `relative_path()`.

src/opi/input/structures/structure_file.py
```diff
@@ -52,17 +54,28 @@ def file(self, val: Path | str | os.PathLike[str]) -> None:
         """
         self._file = Path(val).expanduser().resolve(strict=True)
 
-    def format_orca(self, *, full_path: bool = False) -> str:
+    def format_orca(self, working_dir: Path, /, *, force_full_path: bool = False) -> str:
```
timmyte: I would suggest overhauling the API here completely, as it's not consistent anymore:
```
def format_orca(self, working_dir: Path | None, *, absolute_path: bool = False, original_path: bool = False) -> None:
    # absolute_path is only a suggestion. 
    # You can also omit it.
    if abolsute_path:
        filename = self.file.resolve().expanduser()
    elif original_path:
        filename = self.file
    elif working_dir is not None:
        filename = self.file.relative_to(working_dir)
    else:
        filename = self.file.name

File: src/opi/input/structures/structure_file.py [hunk 3]
```diff
@@ -52,17 +54,28 @@ def file(self, val: Path | str | os.PathLike[str]) -> None:
...
         """
-        Format respectively line in ORCA input.
+        Format respectively line in ORCA input. If the file lies within the working directory a relative path is
+        used, but if it lies outside an absolute path is employed.
```
timmyte: "but if it lies outside an absolute path is employed."
No. The path is never resolved here. Furthermore, by default, `force_full_path` is off. The original name `full_path` is misleading anyway.

src/opi/utils/misc.py
```diff
@@ -173,3 +173,22 @@ def check_minimal_version(version: OrcaVersion, /) -> bool:
     version : OrcaVersion
     """
     return cast(bool, version >= ORCA_MINIMAL_VERSION)
+
+
+def check_file_path(file_path: Path, working_dir: Path) -> Path:
```
timmyte: See my rename suggestion above.

src/opi/utils/misc.py
```diff
@@ -173,3 +173,22 @@ def check_minimal_version(version: OrcaVersion, /) -> bool:
...
+    Returns the relative path to a file and checks if the file exists.
+
+    Raises
+    ------
+    ValueError
+        If relative path cannot be resolved.
+    FileNotFoundError
+        If the file does not exist.
+    """
+    rel = file_path.relative_to(working_dir, walk_up=True)
```
timmyte: Be careful with using: `walk_up`
This was introduced with Python 3.12. But OPI states to be compatible with 3.11. So we cannot use this parameter here.

haneug:
First, I renamed `check_file_path()` to `relative_path()` but then I decided to remove it since I do not really see the advantage of it.

src/opi/core.py
```diff
@@ -214,7 +213,7 @@ def write_input(self) -> None:
                 if (ncores := input_param.ncores) is not None:
                     inp.write(f"%pal\n    nprocs {ncores:d}\nend\n")
                 if (moinp := input_param.moinp) is not None:
-                    inp.write(f'%moinp "{relative_path(moinp, self.working_dir)}"\n')
+                    inp.write(f'%moinp "{moinp.relative_to(self.working_dir)}"\n')
```
timmyte, line 218: Please document that this method might raise `ValueError` before you merge.
As `ValueError` will not be caught by `IOError`.

PR #55, opened 2025-07-10
Extended functionalities for external methods

cplett:
This update adds functionality to improve handling of wrapper scripts for energy and gradient calculations. Key additions include:
1. A standardized interface for ORCA input and output handling to facilitate communication with wrapper scripts.
2. Process management utilities for launching and controlling external wrappers.
3. Support for script-based servers that handle calculations externally.
4. Support for server interfaces that accept custom calculator objects (e.g., for computing energy and gradients using user-defined methods).


src/opi/external_methods/__init__.py
```diff
@@ -0,0 +1,18 @@
+"""
+This sub-package contains all modules and classes required for using an external method for energies and gradients.
+Those are namely:
+    * `process`: Class for managing a python subprocess
```
timmyte: Stick to correct casing. If you mean the class, start with an upper letter.

src/opi/external_methods/interface.py
```diff
@@ -0,0 +1,107 @@
...
+
+
+class ExtoptInterface:
+    """
+    Class for reading and writing ORCA files related to the ExtOpt keyword
+    """
+
+    def read_extopt_input(
+        self, filepath: str
+    ) -> Tuple[str, int, int, int, bool, Optional[str]]:
```
timmyte: This is the old style variant. 
Modern variant: `tuple[str, int, int, int, bool, str | None]:`

src/opi/external_methods/interface.py
```diff
@@ -0,0 +1,107 @@
+from typing import Tuple, Optional
+
+
+class ExtoptInterface:
+    """
+    Class for reading and writing ORCA files related to the ExtOpt keyword
+    """
+
+    def read_extopt_input(
+        self, filepath: str
```
timmyte: File paths should be of type `Path` not plain string.
Or `Path | str`, but then the string should be converted to `Path` ASAP.

src/opi/external_methods/interface.py
```diff
@@ -0,0 +1,107 @@
...
+    ) -> Tuple[str, int, int, int, bool, Optional[str]]:
+        """
+        Reads an input file written by ORCA and returns the parsed values as a tuple.
+
+        Attributes
+        ----------
+        filepath: str
```
timmyte: See above.

src/opi/external_methods/interface.py
```diff
@@ -0,0 +1,107 @@
...
+            Path to the EXT input file.
+
+        Returns
+        -------
+        Tuple containing:
```
timmyte: FYI: The tuple is implied.

src/opi/external_methods/interface.py
```diff
@@ -0,0 +1,107 @@
...
+            xyz_filename: str
+            charge: int
+            multiplicity: int
+            number of cores: int
+            do_gradient: bool
+            pointcharge_filename: str | None, optional
```
timmyte: This return value is not optional.
Optional would mean in some case the function would only return 5 values rather than 6.

src/opi/external_methods/interface.py
```diff
@@ -0,0 +1,107 @@
...
+            charge,
+            multiplicity,
+            ncores,
+            do_gradient,
+            pointcharge_filename,
+        )
+
+    def write_orca_input(
+        self,
+        filename: str,
```
timmyte: See above about `Path`.

src/opi/external_methods/interface.py
```diff
@@ -0,0 +1,107 @@
...
+        nat: int,
+        etot: float,
+        grad: list[float] = None,
+    ) -> None:
+        """
+        Writes an input for ORCA similar to external-tools format.
+
+        Attributes
+        ----------
+        filepath: str
```
timmyte: See above.

src/opi/external_methods/process.py
```diff
@@ -0,0 +1,56 @@
...
+    """
+
+    def __init__(self):
+        """
+        Init Process object
+        """
+        # Variable for storing the process
+        self.process = None
+
+    def start(self, cmd: list[str], pipe=False) -> None:
```
timmyte: Type annotation for `pipe` missing.

src/opi/external_methods/process.py
```diff
@@ -0,0 +1,56 @@
...
+        """
+        Starts any process
```
timmyte: Really **any** process?

src/opi/external_methods/process.py
```diff
@@ -0,0 +1,56 @@
...
+
+        Attributes
+        ----------
+        cmd: str
+            List of arguments to be executed
+        pipe: bool, default: False
+            Determines whether to hold the channel open for piping input in or not.
+        """
```
timmyte, line 54: Please add a "Raises" block mentioning that `subprocess.Popen` might raise `OSError`, `ValueError` or `subprocess.SubprocessError`.

src/opi/external_methods/process.py
```diff
@@ -0,0 +1,56 @@
...
+        if self.process is None or self.process.poll() is not None:
```
timmyte: Why not use `process_is_running()` here!?

src/opi/external_methods/process.py
```diff
@@ -0,0 +1,56 @@
...
+            print("Process is already running.")
+
+    def stop_process(self) -> None:
+        """
+        Kills the process if its running.
+        """
+        if self.process and self.process.poll() is None:
+            print(f"Stopping process with PID {self.process.pid}")
+            self.process.terminate()
+            self.process.wait()
```
timmyte: This can deadlock. Better to first soft kill and after a waiting time hard kill.
```
# > Soft kill
self.process.terminate()
try:
    # > Wait 30sec
    self.process.wait(timeout=30)
except subprocess.TimeoutExpired:
    # > Hard kill
    proc.kill()
finally:
    self.process = None
```

src/opi/external_methods/server.py
```diff
@@ -0,0 +1,178 @@
...
+        self._port = port
+
+    def start_server(self) -> None:
+        """
+        Starts the Server from script
+        Provides _host_id and _port via command line arguments
+        """
+        # First check, whether server.port is free
+        if self.server_port_in_use():
+            print(
```
timmyte: I would kill the process and return a specific return code, so the owner of the server process can determine why the server failed.

src/opi/external_methods/server.py
```diff
@@ -0,0 +1,178 @@
...
+                f"Couldn't setup server on port {self._port} as the port is already in use."
+            )
+            print("Please check whether all previous server were terminated.")
+        else:
+            # Start server by running a python process
+            # Therefore, first set up the command line call for the server script
+            if self.process.process is None or self.process.poll() is not None:
```
timmyte: What about using `process_is_running()`?

src/opi/external_methods/server.py
```diff
@@ -0,0 +1,178 @@
...
+                cmd.append("--port")
+                cmd.append(str(self._port))
+                # Start the server
+                self.process.start(cmd)
+                print(f"Server started with PID {self.process.process}")
+            else:
+                print("Server is already running.")
+
+            # Wait a little to make sure server is fully initialized
+            time.sleep(5)
```
timmyte: Nice catch!

src/opi/external_methods/server.py
```diff
@@ -0,0 +1,178 @@
...
+
+    def kill_server(self):
```
timmyte: Docstring missing.

src/opi/external_methods/server.py
```diff
@@ -0,0 +1,178 @@
...
+import pickle
+import socket
+import time
+
+from opi.external_methods.process import Process
+
+
+class OpiServer:
+    """
+    Class for running a server from a file via socket.
```
timmyte: I assume you mean "network socket"!?

src/opi/external_methods/server.py
```diff
@@ -0,0 +1,178 @@
...
+        port: int
+            Port to use.
+        """
+        self._host_id = host_id
+        self._port = port
+
+    def start_server(self) -> None:
+        """
+        Starts the Server from script
+        Provides _host_id and _port via command line arguments
```
timmyte: Provides `_host_id` and `_port` via command line arguments

src/opi/external_methods/server.py
```diff
@@ -0,0 +1,178 @@
...
+            else:
+                print("Server is already running.")
+
+            # Wait a little to make sure server is fully initialized
+            time.sleep(5)
+
+    def kill_server(self):
+        self.process.stop_process()
+
+    def server_port_in_use(self) -> bool:
```
timmyte, line 151: Suggestion: This function could go into the `misc` package. As it's very general.

src/opi/external_methods/server.py
```diff
@@ -0,0 +1,178 @@
...
+        return False  # Port is free
+
+
+class CalcServer:
+    """
+    Class for running an OpiServer and load a calculator via pickle
+    """
+
+    def __init__(
+        self, serverpath: str, calculator=None, host_id="127.0.0.1", port=9000
```
timmyte: `serverpath: Path, calculator: ??? = None, host_id: ??? = 127.0.0.1`

src/opi/external_methods/server.py
```diff
@@ -0,0 +1,178 @@
...
+    ):
+        """
+        Initialize server object with default values.
+
+        Attributes
+        ----------
+        serverpath: str
+            Path to the server script
+        calculator: obj, default: None
+            A calculator that could be send to the server
```
timmyte: Does it have to have any specific properties!?
Or can literally send, any object!?

src/opi/external_methods/server.py
```diff
@@ -0,0 +1,178 @@
...
+        Change the host_id and port if server not running.
+
+        Attributes
+        ----------
+        host_id: str
+            host_id ID.
+        port: int
+            Port to use.
+        """
+        if not self.server.server_port_in_use():
```
timmyte, line 221: How does the caller of `set_id_and_port()` know if the IP and the port could be assigned or not!?

src/opi/external_methods/server.py
```diff
@@ -0,0 +1,178 @@
...
+            self.server.set_id_and_port(host_id=host_id, port=port)
+
+    def load_calculator(self) -> None:
+        """
+        Send the calculator to the server via pickle.
+        """
+        # Open a connection to server
+        with socket.create_connection(
```
timmyte: Python's `multiprocessing` library has "Client and Listeners": https://docs.python.org/3/library/multiprocessing.html#module-multiprocessing.connection

This would reduce the boilerplate here.

src/opi/external_methods/server.py
cplett: It can be any object. However, it is intended to be a calculator for computing the electronic energy and gradients.

src/opi/external_methods/server.py
cplett: I think it is better to keep it in the Server class to have the function directly accessible via the respective object.

src/opi/external_methods/process.py
```diff
@@ -0,0 +1,78 @@
...
+        Starts a subprocess
+
+        Attributes
+        ----------
+        cmd: str
+            List of arguments to be executed
+        pipe: bool, default: False
+            Determines whether to hold the channel open for piping input in or not.
+
+        Raises
```
timmyte, line 42: The purpose of the "Raises" block is to state which exceptions a function might throw when called. But as you catch all and handle all of these exceptions already, none of these will ever be raised by this function.
I would suggest not catching them and let them propagate naturally.

src/opi/external_methods/process.py
```diff
@@ -0,0 +1,78 @@
...
+        """
+        if self.process_is_running():
+            print(f"Stopping process with PID {self.process.pid}")
+            # Softkill
+            self.process.terminate()
+            try:
+                self.process.wait(timeout=30)
+                self.process = None
+            except subprocess.TimeoutExpired:
+            # Hard kill
```
timmyte: * Indent comment.
* Why not use `finally` here?

src/opi/external_methods/server.py
```diff
@@ -0,0 +1,188 @@
...
+        self._port = port
+
+    def start_server(self, exe: str=sys.executable) -> None:
+        """
+        Starts the Server from script
+        Passes self._host_id and self._port as command-line arguments to the server script.
+
+        Parameters
+        ----------
+        exe: str, default=sys.executable
```
timmyte:
```suggestion
        exe: str, default: sys.executable
```

src/opi/external_methods/server.py
```diff
@@ -0,0 +1,188 @@
...
+        ----------
+        host_id: str
+            host_id ID.
+        port: int
+            Port to use.
+        """
+        self._host_id = host_id
+        self._port = port
+
+    def start_server(self, exe: str=sys.executable) -> None:
```
timmyte:
```suggestion
    def start_server(self, exe: str = sys.executable) -> None:
```

src/opi/external_methods/server.py
```diff
@@ -0,0 +1,188 @@
...
+
+from opi.external_methods.process import Process
+
+
+class OpiServer:
+    """
+    Class for running a server from a file using a network socket.
+    """
+
+    def __init__(self, serverpath: str, host_id="127.0.0.1", port=9000):
```
timmyte:
```suggestion
    def __init__(self, serverpath: str, host_id : str = "127.0.0.1", port: int = 9000):
```

src/opi/external_methods/interface.py
```diff
@@ -0,0 +1,108 @@
...
+        multiplicity = int(lines[2])
+        ncores = int(lines[3])
+
+        # Check if gradient should be calculated or not
+        if int(lines[4]) == 0:
+            do_gradient = False
+        elif int(lines[4]) == 1:
+            do_gradient = True
+        else:
+            raise ValueError("do_gradient from ORCA input must be 0 or 1.")
```
timmyte: Safer variant would be:
```
try:
   do_gradient = int(lines[4]) == 1
except ValueError as err
    raise ValueError("do_gradient from ORCA input must be 0 or 1.") from err
```

src/opi/external_methods/process.py
```diff
@@ -0,0 +1,78 @@
...
+    """
+    Class that runs a process, monitors it and terminates it.
+    """
+
+    def __init__(self):
+        """
+        Init Process object
+        """
+        # Variable for storing the process
+        self.process = None
```
timmyte:
```suggestion
        self.process: subprocess.Popen | None = None
```

src/opi/external_methods/server.py
```diff
@@ -0,0 +1,188 @@
...
+                f"Couldn't setup server on port {self._port} as the port is already in use."
+            )
+            print("Please check whether all previous server were terminated.")
+            sys.exit(101) # Exit and return 101 code
+        else:
+            # Start server by running a python process
+            # Therefore, first set up the command line call for the server script
+            if not self.process.process_is_running():
+                # Build the command list:
+                # ["python", server_script] + --host_id ID + --port port + optional args
```
timmyte: The command assemble below does not have the `--host_id` and `--port` flags.
So what is the correct format of the command?

src/opi/external_methods/server.py
timmyte: One could have a method here that wraps the original function in `misc` ;)
[Not a must-do!]

src/opi/external_methods/server.py
```diff
@@ -0,0 +1,188 @@
...
+        return False  # Port is free
+
+
+class CalcServer:
+    """
+    Class for running an OpiServer and load a calculator via pickle
+    """
+
+    def __init__(
+        self, serverpath: str, calculator: Any=None, host_id: str="127.0.0.1", port: int=9000
```
timmyte:
```suggestion
        self, serverpath: str, calculator: Any = None, host_id: str = "127.0.0.1", port: int = 9000
```

src/opi/external_methods/server.py
```diff
@@ -0,0 +1,188 @@
...
+        Send the calculator to the server via pickle.
+        """
+        # Open a connection to server
+        address = (self.server._host_id, self.server._port)
+        with Client(address, authkey=b'secret') as conn:
+            conn.send({"type": "setup_calculator", "calculator": self._calculator})
+            response = conn.recv()
+            print("Server setup response:", response)
+
+    def start_server(self, exe: str=sys.executable) -> None:
```
timmyte:
```suggestion
    def start_server(self, exe: str = sys.executable) -> None:
```

src/opi/external_methods/server.py
timmyte: What I meant is: I would return a boolean.

src/opi/external_methods/server.py
```diff
@@ -0,0 +1,188 @@
...
+        else:
+            print("ID and port could not be set up.")
+
+    def load_calculator(self) -> None:
+        """
+        Send the calculator to the server via pickle.
+        """
+        # Open a connection to server
+        address = (self.server._host_id, self.server._port)
+        with Client(address, authkey=b'secret') as conn:
```
timmyte: Why use an `authkey` in the first place?
Encryption and decryption takes time.

src/opi/external_methods/server.py
```diff
@@ -0,0 +1,188 @@
...
+        exe: str, default=sys.executable
+            Executable to use for starting the server
+        """
+        # First check, whether server.port is free
+        if self.server_port_in_use():
+            print(
+                f"Couldn't setup server on port {self._port} as the port is already in use."
+            )
+            print("Please check whether all previous server were terminated.")
+            sys.exit(101) # Exit and return 101 code
```
timmyte: Use an `IntEnum` and make sure that the routine in OPI that calls the server also has access to that enum. As OPI has to react to failing server at the end!

timmyte (changes requested):
All the printing of errors is nice for logging. But also consider batch usage.
Multiple OPI processes are started, each running its own server. 
The only message the user sees in their log in the end are, e.g.:
```
ID and port could not be set up.
ID and port could not be set up.
ID and port could not be set up.
ID and port could not be set up.
ID and port could not be set up.
ID and port could not be set up.
ID and port could not be set up.
[...]
```
But this still leaves open questions:
* Do all these error messages come from the same server process?
* Did all processes fail?

The classes that you design here are supposed to be controlled by other program parts and not by humans directly. 
So they should return success and error messages in a computer processable way.

src/opi/external_methods/process.py
cplett: If I'm not mistaken, the process is always tried to be hard killed even if the softkill succeeded. One might do this, but then the softkill is probably not required.

src/opi/external_methods/process.py
```diff
@@ -0,0 +1,111 @@
...
+from typing import Any
+
+
+class Process:
+    """
+    Class that runs a process, monitors it and terminates it.
+    """
+
+    # Error type definitions
+    class ProcessAlreadyRunningError(Exception):
```
timmyte: Don't nest classes. This is bad style.

timmyte (changes requested):

timmyte (changes requested):
Please run `nox` and commit changes!

timmyte (approved):

PR #56, opened 2025-07-10
Add get_ for molecular orbitals and restructure available getters.

haneug:
Adds several new get functions:
- `get_hftype()` with the corresponding enum to identify rhf/uhf/casscf calculations.
- `get_charge()` and `get_mult()` for molecular charge and multiplicity.
- `get_electrons()` for the number of electrons, optional alpha/beta resolved.
- `get_gradient()` for easier access to the nuclear gradient.
- `get_mos()` returns a dict with list(s) of MOs, with key "mos" for rhf or "alpha"/"beta" for uhf.
- `get_homo()` returns the HOMO.
- `get_lumo()` returns the LUMO.
- `get_hl_gap()` calculates the HOMO-LUMO gap.
- `get_structure()` now does not throw an error but also returns None if no structure is available. This is done to be consistent with the other get functions.
- reordered the get functions

### DIFF
--- a/src/opi/output/core.py
+++ b/src/opi/output/core.py
@@ -484,11 +484,16 @@ class Output:
 
         Returns
         -------
-        nel, None | alpha, beta | None, None: tuple[int, int] | tuple[None,None]
-            Returns the number of electrons (spin resolved if requested) as integers or None:
-                - nel, None, all electrons, not spin resolved.
-                - nalpha, nbeta, alpha and beta electrons separated.
-                - None, None, if no electrons could be retrieved, or no multiplicity for spin resolution is available.
+        nalpha : int | None
+            * None: if no electrons could be retrieved, or no multiplicity for spin resolution is available.
+            * int:
+                * if `spin_resolved == False`, the total number of electrons
+                * if `spin_resolved == True`, the number of alpha electrons.
+        nbeta : int | None
+            * None:
+                * if no electrons could be retrieved, or no multiplicity for spin resolution is available.
+                * if `spin_resolved == False`
+            * int : number of beta electrons (requires `spin_resolved == True`)
         """
         nel = self._safe_get("results_properties", "calculation_info", "numofelectrons")
 
@@ -507,11 +512,11 @@ class Output:
             return None, None
 
         # > Calculate amount of alpha and beta electrons
-        alpha = nel // 2 + (mult - 1)
-        beta = nel - alpha
+        nalpha = nel // 2 + (mult - 1)
+        nbeta = nel - nalpha
 
         # > Return spin resolved number of electrons
-        return alpha, beta
+        return nalpha, nbeta
 
     def get_final_energy(self, *, index: int = -1) -> StrictFiniteFloat | None:
         """


examples/exmp003_opt/job.py
```diff
@@ -66,4 +66,7 @@
 
     # > Now we print the final structure as xyz file
     optimized = output.get_structure()
-    print(optimized.to_xyz_block())
\ No newline at end of file
+    print(optimized.to_xyz_block())
+
+    # > Now we print the final gradient
+    print(output.get_gradient(index=-2))
```
timmyte, line 73: If it's the "final" why is the index `-2`?

src/opi/output/core.py
```diff
@@ -419,113 +421,92 @@ def _safe_get(self, *attrs: str | int) -> Any | None:
...
-                list[tuple[StrictStr, StrictFiniteFloat, StrictFiniteFloat, StrictFiniteFloat]],
-                cartesians,
-            )
+        hftype = self._safe_get("results_gbw", "molecule", "hftyp")
+        if hftype is not None:
+            hftype = cast(str, hftype)
+            try:
+                hftype = Hftypes(hftype)
+            except ValueError:
+                hftype = Hftypes.UNKNOWN
```
timmyte: In what situations would the HFType be not known?
Why not just return `None` here?

src/opi/output/core.py
```diff
@@ -419,113 +421,92 @@ def _safe_get(self, *attrs: str | int) -> Any | None:
...
-        Raises
-        ----------
-        ValueError
-            If no geometry with the requested index is available.
+        if mult is not None:
+            mult = cast(int, mult)
+
+        return mult
+
+    def get_electrons(self, *, spin_resolved: bool = False) -> int | tuple[int, int] | None:
```
timmyte: Suggestions:
* I would rename the function `get_nelectrons()` or `get_num_electrons()` to emphasize that it returns integers.
*  IMO changing the type of the return value(s) based on input parameters is bad style. Therefore, I would suggest opting for:
```
-> tuple[None, None] | tuple[int, int | None]
```
Where `tuple[None, None]` is returned instead of `None`.
And in any other case `tuple[int, int | None]`. 
So the caller always has two values to unpack:
```
nelec_a, nelec_b = get_electrons(...)
```

examples/exmp003_opt/job.py
haneug: Good catch. It is the final calculated gradient. For the final structure no gradient is calculated by ORCA and None will be found. I will update the documentation of the function accordingly.

src/opi/output/core.py
haneug: I am not sure when this would apply. But None is also fine.

src/opi/output/core.py
haneug: I agree and adjusted it accordingly

examples/exmp003_opt/job.py
timmyte: Thanks!

src/opi/output/core.py
```diff
@@ -419,113 +421,97 @@ def _safe_get(self, *attrs: str | int) -> Any | None:
...
-        ----------
-        structure: Structure
-            Returns structure object generated from the output for the given index.
-
-        Raises
-        ----------
-        ValueError
-            If no geometry with the requested index is available.
+        -------
+        nel, None | alpha, beta | None, None: tuple[int, int] | tuple[None,None]
```
timmyte: I find this description very confusing. How about something like:
```
Returns
-------
nalpha : int | None
    * None: if no electrons could be retrieved, or no multiplicity for spin resolution is available.
    * int: 
       * if `spin_resolved == False`, the total number of electrons
       * if `spin_resolved == True`, the number of alpha electrons.
nbeta : int | None
    * None: 
       * if no electrons could be retrieved [...]
       * if `spin_resolved == False`
    * int : number of beta electrons (requires `spin_resolved == True`)
```

haneug: I agree and updated the docstring accordingly.

timmyte (approved):

PR #60, opened 2025-07-14
Initial support for orca_plot

haneug:
- Initial support for plotting MOs with orca_plot as part of the output class.
- Includes an example how to visualize MOs in the notebooks

@timmyte I would highly appreciate your feedback here if this approach is okay. One danger of calling orca_plot would be that it runs forever without doing anything so instead of the interactive mode we could also write an input file for orca_plot and avoid this. However, I believe as soon as something is not present orca_plot will exit and the input file itself is annoying. 

**To Add:**
- [x] MO by index
- [x] Density
- [x] Spin density

### DIFF
--- a/src/opi/output/cube.py
+++ b/src/opi/output/cube.py
@@ -24,6 +24,11 @@ class CubeOutput:
         """
         Reads the cube data from file at stored path and returns it as string.
 
+        Returns
+        ----------
+        str
+            Returns a string with the cube data.
+
         Raises
         ----------
         FileNotFoundError
@@ -31,16 +36,21 @@ class CubeOutput:
         """
         return self._path.read_text()
 
-    def iter_cube(self) -> Iterator[str]:
+    def __iter__(self) -> Iterator[str]:
         """
         Lazily yields lines from the cube file (memory efficient).
-        """
-        with self._path.open("r") as f:
-            for line in f:
-                yield line
 
-    def __iter__(self) -> Iterator[str]:
-        return self.iter_cube()
+        Returns
+        ----------
+        Iterator[str]
+            Iterator that yields the cube data line by line.
+
+        Raises
+        ----------
+        FileNotFoundError
+            If the cube file does not exist.
+        """
+        return iter(self._path.open())
 
     def __str__(self) -> str:
         """Returns the name of the class and the path the object holds"""


src/opi/execution/core.py
```diff
@@ -163,6 +164,8 @@ def run(
             Name of ORCA binary to be executed. Path is automatically resolved based on configuration.
         args : Sequence[str], default: ()
             Command line arguments to pass to ORCA binary.
+        input_string: str | None = None
```
timmyte: IMO, `stdin_str` would be a more appropriate name.

src/opi/execution/core.py
```diff
@@ -285,7 +289,59 @@ def run_orca(
...
+        # Sets the output and error file from the gbwfile.
+        outfile = gbwfile.with_suffix(".plot.out")
+        errfile = gbwfile.with_suffix(".plot.err")
+
+        # > CLI arguments
+        arguments = [gbwfile.name]
+        # > Request interactive plot mode by adding "-i"
+        arguments += ["-i"]
+        if extra_args:
+            # > All extra arguments are passed as second argument to orca_plot.
```
timmyte:
```suggestion
            # > All extra arguments are passed as third argument to orca_plot.
```

src/opi/execution/core.py
```diff
@@ -285,7 +289,59 @@ def run_orca(
...
+            arguments += [shlex.join(extra_args)]
```
timmyte: Question: Why do you throw in all extra arguments as one and don't just append them to the list?

src/opi/execution/core.py
```diff
@@ -285,7 +289,59 @@ def run_orca(
...
+        self,
+        gbwfile: Path,
+        input_string: str,
+        /,
+        *extra_args: str,
+        silent: bool = True,
+        timeout: int = -1,
+    ) -> None:
+        """
+        Executes the orca_plot binary and passes the gbw path, an input string, and extra arguments to the binary.
```
timmyte: Please mention that this function only supports `orca_plot <gbw file> -i ...` currently and not any of the other usage variants.

src/opi/execution/core.py
```diff
@@ -285,7 +289,59 @@ def run_orca(
...
+
+        Parameters
+        ----------
+        gbwfile : Path
+            Path to an ORCA geometry, basis set, wavefunction (gbw) file.
+        input_string : str
```
timmyte: * Did you consider that this string requires newline characters between every token? Taking a list here using `'\n'.join(...) + '\n'` could be safer and more user-friendly.
* Also take into account that if the sequence is faulty, the interactive program will probably deadlock, as it's waiting for input. Consider having a positive `timeout` here by default. Maybe 60 seconds or 300 seconds?

src/opi/execution/core.py
```diff
@@ -285,7 +289,59 @@ def run_orca(
...
+            Input string handed to stdin of orca_plot.
+        *extra_args: str
+            Additional arguments passed to orca_plot.
+        silent : bool, default: True
+            Capture and discard STDOUT and STDERR.
+        timeout : int, default: -1
+            Optional timeout in seconds to wait for process to complete.
+        """
+        if not gbwfile.is_file():
```
timmyte, line 331: Also check that `input_string` is not an empty string and that it ends with a newline character, otherwise the process deadlocks.

src/opi/output/core.py
```diff
@@ -376,6 +380,91 @@ def geometry_optimization_converged(self) -> bool:
...
+        Raises
+        ----------
+        ValueError
+            If the input string is empty a ValueError is raised.
+        """
+        runner = self._create_runner()
+        gbwfile = self.working_dir / f"{self.basename}{suffix}"
+
+        if not input_string:
+            raise ValueError
```
timmyte: Add some text to the exception please.

src/opi/output/core.py
```diff
@@ -376,6 +380,91 @@ def geometry_optimization_converged(self) -> bool:
...
+            Input string for interactive orca_plot session
+        suffix : str, default ".gbw"
+            Determines the suffix of the gbw file to use.
+        timeout : int, default: = -1
+            Timeout in seconds to wait for ORCA process.
+            If value is smaller than zero, wait indefinitely.
+
+        Raises
+        ----------
+        ValueError
```
timmyte, line 404: Are these all exceptions this routine could raise? What about `FileNotFoundError`?

src/opi/output/core.py
```diff
@@ -376,6 +380,91 @@ def geometry_optimization_converged(self) -> bool:
...
+        operator : StrictNonNegativeInt, default=0
+            Operator of the MO, alpha MOs are indicated by 0 and beta MOs by 1.
+        resolution: StrictNonNegativeInt, default=40
+            Resolution of the generated cube file. Higher numbers result in smoother plots, but also in longer orca_plot
+            run time and a larger cube file.
+
+        Returns
+        -------
+        str | None
+            Returns the cube file as string or returns None if the cube file cannot be retrieved.
```
timmyte: Correct me if I'm wrong, but cube files can be huge.
I wouldn't just pass around their contents as plain string. I would return either the file path, so the caller can decide what to do with the data, or return a generator.

src/opi/execution/core.py
haneug: I append "\n12\n" to the string which will terminate orca_plot with an empty string. Is that sufficient?

src/opi/output/core.py
haneug: You are right, I added `FileNotFoundError` to the Raises

src/opi/execution/core.py
haneug: I updated the name to `stdin_str`.

src/opi/execution/core.py
haneug: - I now take a list and join it in run_orca_plot.
- That is a fair point but the thing is that orca_plot can potentially take relatively long. If the system is large and the grid is fine, runtime can easily exceed 300 seconds. I decided to add the timeouts to the higher level functions the users will actually use and warn them about this in the docstring. This also allows different default timeouts for different plot types, e.g., mos take less time than densities.

src/opi/output/core.py
haneug: I agree but I want to give easy access to the string for plotting cube data in the notebooks.  My solution is to add a CubeOutput object that keeps the path and reads the file when the property is requested.

src/opi/execution/core.py
haneug: Good point. I append them now.

src/opi/execution/core.py
haneug: Nevermind, I now check if the list is provided and due to the list approach it will be terminated with \n

src/opi/execution/core.py
```diff
@@ -163,6 +163,8 @@ def run(
             Name of ORCA binary to be executed. Path is automatically resolved based on configuration.
         args : Sequence[str], default: ()
             Command line arguments to pass to ORCA binary.
+        std_str: str | None = None
```
timmyte:
```suggestion
        stdin_str: str | None = None
```

src/opi/output/cube.py
```diff
@@ -0,0 +1,34 @@
...
+
+class CubeOutput:
+    """
+    Class that stores the path to a cube file and provides easy access to the cube
+    data via the cube property.
+    """
+
+    def __init__(self, path: Path):
+        if path.is_file():
+            self._path = path
```
timmyte, line 13: I guess, this parameter does not need to be hidden.
If it's altered, no other data is compromised.

src/opi/output/cube.py
```diff
@@ -0,0 +1,34 @@
...
+        Reads the cube data from file and returns it as string.
+
+        Raises
+        ----------
+        FileNotFoundError
+            If the cube file does not exist anymore.
+        """
+        return self._path.read_text()
+
+    def __str__(self) -> str:
```
timmyte, line 55: I wouldn't define the `__str__()` like this. That's lots of data that would flood the terminal. I would rather return something like: `f"{self.__class__.name}({self._path})"`, which should look as follows, e.g.: `CubeOutput(/home/test/file.cube)`

Stick to the `cube` property to return the actual data.

src/opi/output/cube.py
```diff
@@ -0,0 +1,34 @@
+from pathlib import Path
+
+
+class CubeOutput:
```
timmyte, line 5: Nice idea ;)
You could think about adding the `__iter__()` method, allowing lazy access to the data. For "big data" it's always a good idea to also allow lazy access.

timmyte (changes requested):

src/opi/output/cube.py
haneug: I would prefer to keep it hidden since I check that the file exists upon initialization and I would like to prevent the user from changing the path.

src/opi/output/cube.py
```diff
@@ -0,0 +1,47 @@
...
+        Reads the cube data from file at stored path and returns it as string.
+
+        Raises
+        ----------
+        FileNotFoundError
+            If the cube file does not exist.
+        """
+        return self._path.read_text()
+
+    def iter_cube(self) -> Iterator[str]:
```
timmyte: 1) Please remove `iter_cube()` and just put the functionality directly into `__iter__()`. Defining `__iter__()` as a thin wrapper around `iter_cube()` just adds unnecessary overhead. 
2) Don't forget to also document `FileNotFoundError` as this could also be raised.

src/opi/output/cube.py
```diff
@@ -0,0 +1,47 @@
...
+        """
+        Lazily yields lines from the cube file (memory efficient).
+        """
+        with self._path.open("r") as f:
+            for line in f:
+                yield line
+
+    def __iter__(self) -> Iterator[str]:
+        return self.iter_cube()
```
timmyte:
```suggestion
    def __iter__(self) -> Iterator[str]:
        return iter(self._path.open())
```
Et voila ;)

PR #61, opened 2025-07-16
Arbitrary block variable

nakul680:
- Allows user to add an arbitrary block variable , which will be printed in block along with available variables.
-  Arbitrary variables are stored in a dict which users can add to or remove from or even clear completely.


src/opi/input/blocks/base.py
```diff
@@ -16,18 +16,59 @@ class Block(BaseModel, ABC):
...
     Every class defined for a block is derived from this base Block class ,
     which defines attributes, methods and properties shared by all blocks.
+
+    Attributes
+    ----------
+        aftercoord (bool): Indicates whether the block is positioned after a coordinate transformation.
+
+    Private Attributes
+    ------------------
+        _name (str): Internal name identifier for the block.
```
haneug: Please use proper numpy style docstrings and check the generated autoapi!

haneug (changes requested):
Nice addition, please adjust the docstrings and check the autoapi!

src/opi/input/blocks/base.py
```diff
@@ -16,18 +16,91 @@ class Block(BaseModel, ABC):
...
     Every class defined for a block is derived from this base Block class ,
     which defines attributes, methods and properties shared by all blocks.
+
+    Attributes
+    ----------
+        aftercoord: bool
+            Indicates whether the block is positioned after a coordinate transformation.
+        _name | name: str
+            Internal name identifier for the block.
+        _arbitrary | arbitrary: dict[str, str]
```
timmyte: This class has no attribute `abitrary` only `_arbitrary`.

src/opi/input/blocks/base.py
```diff
@@ -16,18 +16,91 @@ class Block(BaseModel, ABC):
...
+            A dictionary storing arbitrary variable names as keys and the variable values as value. Both are stored as strings.
```
timmyte: Be more precise. The dictionary is not meant to just store anything, but key-value options for the ORCA input, that are not implemented natively!

src/opi/input/blocks/base.py
```diff
@@ -16,18 +16,91 @@ class Block(BaseModel, ABC):
...
     """
 
     model_config = ConfigDict(arbitrary_types_allowed=True)
     _name: str
     aftercoord: bool = False
+    _arbitrary: dict[str, str] = {}
+
+    def add_arbitrary_attributes(self, var: dict[str, str]) -> None:
```
timmyte: 1) I would label this functionality `add_option()`.
Analogously, for the `remove`, `clear`, `has` methods
2) We need to be stricter here:
   * This method should take `name: str` and a `value: str`.
   * The `name` has to be converted to lowercase.
       * This applies to every method that takes `name`. Each needs to perform a case-insensitive look up. How about a custom dict class:
```
from collections import UserDict
from typing import Any

class NoCaseDict(UserDict):
    """
    A dictionary whose keys are case-insensitive.
    This only applies for string-like keys.
    """

    def __getitem__(self, key: Any) -> Any:
        key = self._norm_key(key)
        return self.data[key]

    def __setitem__(self, key: Any, value: Any) -> None:
        key = self._norm_key(key)
        self.data.update({key: value})

    def __delitem__(self, key: Any) -> None:
        key = self._norm_key(key)
        del self.data[key]

    def _norm_key(self, key: Any, /) -> Any:
        try:
            return key.lowercase()
        except AttributeError:
            return key
```
   * Don't touch the `value`.
   * Make sure both are of type `str`.
   * What if an attribute with the same name already exists!?

src/opi/input/blocks/base.py
```diff
@@ -16,18 +16,91 @@ class Block(BaseModel, ABC):
...
+        Add arbitrary attributes to this block.
+
+        Parameters
+        ----------
+        var: dict[str, str]
+            Dictionary storing arbitrary variable names as keys and the variable values as value
+        """
+        self._arbitrary.update(var)
+
+    def remove_arbitrary_attribute(self, name: str) -> None:
```
timmyte: How does this method behave if `name` is not present?
Please document this.

timmyte (changes requested):

haneug:
@nakul680 please also update the number of the example and update before the next review

src/opi/input/blocks/base.py
nakul680: Previously we had discussed if the user overwrites an attribute by using the same name , its an error that we push on to the user . Should I instead add a check if such an attribute already exists and add a force parameter to allow the user to overwrite ?

src/opi/input/blocks/block_scf.py
```diff
@@ -493,6 +493,8 @@ def path_from_string(cls, path: str | InputFilePath) -> InputFilePath:
 
     def format_orca(self) -> str:
```
haneug, line 494: Why do we copy `format_orca` here? If the key is `scflambda` we replace it with `lambda` but could we not just change the key to lambda? Or am I missing something here?

src/opi/input/blocks/util.py
```diff
@@ -181,3 +182,138 @@ def init(cls, inp: str | list[int | tuple[int, int]] | list[int]) -> Self:
...
+        TypeError
+            If the key is not a string.
+
+        """
+        if not isinstance(key, str):
+            raise TypeError(f"Key must be of type string, got {type(key)}")
+        key = self._norm_key(key)
+        del self.data[key]
+
+    def __contains__(self, key: Any) -> bool:
```
haneug, line 272: key should be str

src/opi/input/blocks/util.py
```diff
@@ -181,3 +182,138 @@ def init(cls, inp: str | list[int | tuple[int, int]] | list[int]) -> Self:
...
+        TypeError
+            If the key is not a string.
+
+        """
+        if not isinstance(key, str):
+            raise TypeError(f"Key must be of type string, got {type(key)}")
+        key = self._norm_key(key)
+        return key in self.data
+
+    def _norm_key(self, key: str, /) -> Any:
```
haneug: return type should be str

src/opi/input/blocks/base.py
```diff
@@ -16,18 +16,148 @@ class Block(BaseModel, ABC):
...
+        If the attribute doesn't exist yet, a new attribute will be created.
+
+        Parameters
+        ----------
+        name: str
+            Name of arbitrary attribute.
+        val: str
+            Value of arbitrary attribute.
+
+        Returns
```
haneug: Here you document the `None` return type and for the other functions you do not do this. Maybe stay consistent and remove it here.

src/opi/input/blocks/block_scf.py
haneug: never mind we cannot use lambda. I forgot that.

src/opi/input/blocks/util.py
nakul680: Nox throws an error if the function signature of __contains__ is changed , probably because it is a special method

haneug: Thanks for pointing that out. Then keep it as Any but please update the docstiring to Any as welll

src/opi/input/blocks/util.py
```diff
@@ -181,3 +182,138 @@ def init(cls, inp: str | list[int | tuple[int, int]] | list[int]) -> Self:
...
+
+        Returns
+        -------
+        str
+            Normalized key.
+
+        Raises
+        -------
+        AttributeError
+            If normalization is not possible.
```
timmyte: Are you sure this exception can be raised by this method?

src/opi/input/blocks/util.py
```diff
@@ -181,3 +182,138 @@ def init(cls, inp: str | list[int | tuple[int, int]] | list[int]) -> Self:
...
+
+        Parameters
+        ----------
+        key: Any
+            Key of entry
+
+        Returns
+        -------
+        bool
+            True if key exists in the dictionary.False otherwise.
```
timmyte:
```suggestion
            True if key exists in the dictionary. False otherwise.
```

src/opi/input/blocks/block_scf.py
```diff
@@ -493,6 +493,8 @@ def path_from_string(cls, path: str | InputFilePath) -> InputFilePath:
...
         s = f"%{self.name}\n"
+        for key, value in self._arbitrary.items():
```
timmyte: Please add comments.
Describe what the loops do.

src/opi/input/blocks/base.py
```diff
@@ -16,18 +16,144 @@ class Block(BaseModel, ABC):
...
+        """
+        return str(self._arbitrary.__getitem__(name))
 
     def format_orca(self) -> str:
         """
         Method to convert a Block instance into string for the ORCA input file.
         Returns the string representation of the respective class it is called by.
         """
         s = f"%{self.name}\n"
+        for key, value in self._arbitrary.items():
```
timmyte: Please add comments.
Describe what the loops do.

src/opi/input/blocks/base.py
```diff
@@ -16,18 +16,144 @@ class Block(BaseModel, ABC):
...
+            True if the attribute with the given name exists, False otherwise.
+
+        Raises
+        ------
+        TypeError
+            if name is not of type string
+        """
+        return self._arbitrary.__contains__(name)
+
+    def get_value(self, name: str) -> str | None:
```
timmyte:
```suggestion
    def get_option(self, name: str) -> str | None:
```
Stay in the same namespace.

src/opi/input/blocks/base.py
```diff
@@ -16,18 +16,144 @@ class Block(BaseModel, ABC):
...
+            The value of the attribute if it exists, else None.
+
+        Raises
+        ------
+        KeyError
+            if no attribute with that name exists
+        TypeError
+            if name is not of type string
+        """
+        return str(self._arbitrary.__getitem__(name))
```
timmyte: Below you stated, that `NoCaseDict` only allows strings as values (also in `add_option()`)? Which is correct!
So, why would one have to re-cast a string to a string?

src/opi/input/blocks/util.py
timmyte, line 272: Good catch to add `__contains__()` ;)

src/opi/input/blocks/util.py
```diff
@@ -181,3 +182,138 @@ def init(cls, inp: str | list[int | tuple[int, int]] | list[int]) -> Self:
...
+        """
+        if not isinstance(key, str):
+            raise TypeError(f"Key must be of type string, got {type(key)}")
+        key = self._norm_key(key)
+        return key in self.data
+
+    def _norm_key(self, key: str, /) -> str:
+        """
+        Normalize key to lower case.
+        Parameters
```
timmyte, line 301:
```suggestion
        Normalize key to lower case.
        
        Parameters
```

timmyte (changes requested):

PR #82, opened 2025-07-22
Notebook for visualizing cube data from (localized-)MOs

haneug:
- Introduces a notebook for visualization of (localized)-MOs of Water. First a native GFN2-xTB calculation is performed and the orbitals are localized. Then an MO diagram is plotted and the HOMO-2, HOMO-1, and HOMO orbitals are depicted. Finally the localized orbitals which are linear combinations of the former orbitals are depicted as well.

- I added a new Enum for distinguishing different gbw files containing different orbital types.   

@cplett Please review the notebook, @timmyte please review the code changes.


src/opi/output/gbw_suffixes.py
```diff
@@ -0,0 +1,30 @@
+from opi.models.string_enum import StringEnum
+
+
+class GbwSuffixes(StringEnum):
+    """Enumeration to keep track of the different suffixes of gbw files"""
+
+    gbw = ".gbw"
```
timmyte: By convention, enum members should always be spelled in all uppercase letters.

src/opi/output/gbw_suffixes.py
```diff
@@ -0,0 +1,30 @@
+from opi.models.string_enum import StringEnum
+
+
+class GbwSuffixes(StringEnum):
```
timmyte: By convention, enum names should use a singular type name: `GbwSuffix`

src/opi/output/gbw_suffixes.py
```diff
@@ -0,0 +1,30 @@
...
+
+    @classmethod
+    def from_string(cls, suffix: str) -> "GbwSuffixes":
+        if not suffix.startswith("."):
+            suffix = f"{suffix}"
+        try:
+            return cls(suffix)
+        except ValueError:
+            allowed = ", ".join(m.value for m in cls)
+            raise ValueError(f"Invalid gbw suffix: {suffix!r}. Allowed suffixes: {allowed}")
```
timmyte: Instead of defining a new method here, I would overload the `_missing_()` method,
this would make the API easier: `GbwSuffix(".qro")` instead of `GbwSuffix.from_string(".qro")`

How about something like this. 
```suggestion
from typing import Self
    @classmethod
    def _missing_(cls, value: object, /) -> Self:
        # > Add leading dot is missing
        # >> This allows omitting the dot on construction
        if isinstance(suffix, str) and not suffix.startswith("."):
            suffix = '.' + suffix
        # > Can raise TypeError and ValueError
        return super()._missing_(suffix)
```

timmyte (changes requested):

src/opi/output/core.py
```diff
@@ -441,6 +446,8 @@ def plot_mo(
         timeout: int, default = 300
             Time after which orca_plot will be stopped. 300 seconds should be sufficient for most MOs but when something
             large is plotted set this to a larger value or to -1 for waiting indefinitely long.
+        gbw_type: str | GbwSuffixes, default = GbwSuffixes.GBW
```
```suggestion
        gbw_type: str | GbwSuffix, default = GbwSuffix.GBW
```

PR #83, opened 2025-07-28
Add additional getters, rename output models, support multiple gbw files

haneug:
Adds various getters for the following properties:
- atomic charges
- S² expectation and ideal value from output file 
- thermo corrections
- integrals (H, F, S, J, K)
- dipoles, quadrupoles, polarizabilities

renames some of the output model classes and output model files for consistency. Note that the atomic charges, dipole moments, quadrupole moments, and polarizabilities are just shortcuts to the properties but they do not sort/process them for actual easy use.

**Note:** Also added support for multiple gbw files see comments below


src/opi/execution/core.py
```diff
@@ -567,7 +567,8 @@ def create_gbw_json(
         config: dict[str, bool | str | list[str | int]] | None = None,
     ) -> None:
         """
-        Create the `<basename>.json` file from `<basename>.gbw`.
+        Create the `<basename>.json` file from `<basename>.gbw`. If an index is supplied create the
```
timmyte: The description of the index is unclear.
How does one pass the index?
Does it come as part of the `basename`?
In the scope of this method, does it make a difference if the `basename` ends with an integer or not?

src/opi/output/grepper/recipes.py
```diff
@@ -40,6 +40,48 @@ def has_string_in_file(file_name: Path, search_for: str, /, *, strict: bool = Tr
...
+    file_name : Path
+        Path to the output file
+    search_for : str
+        string that function searches in file.
+    index : int
+        index of occurrence that should be returned.
+    field: int
+        field in line that should be returned
+    strict : bool, default: False
+        True: Raise "FileNotFoundError" exception if `file_name` does not exist.
```
timmyte: I would still add the `Raises` section and just state to look above.

src/opi/output/grepper/recipes.py
```diff
@@ -40,6 +40,48 @@ def has_string_in_file(file_name: Path, search_for: str, /, *, strict: bool = Tr
...
+    """
+    try:
+        grepper = Grepper(file_name)
+        results = grepper.search(
+            search_for,
+            kind=float,
+            field=field,
+            case_sensitive=True,
+        )
+        return float(results[index])
```
timmyte, line 94: What exception will be raised when the casting to float fails?
Please put into the `Raises` block.

src/opi/output/models/json/property/properties/geometries.py
```diff
@@ -0,0 +1,177 @@
...
+)
+
+
+class Geometries(GetItem):
+    """
+    Contains the geometry dependent calculated results in the ORCA job
+
+    Attributes
+    ----------
+    geometry: Geometry | None default = None
```
timmyte:
```suggestion
    geometry: Geometry | None, default = None
```
Please correct for all attributes.

src/opi/output/core.py
```diff
@@ -110,7 +133,9 @@ def __init__(
             raise FileNotFoundError(f"Working dir does not exist: {working_dir}")
 
         # // JSON PATHS
-        self.gbw_json_file = self.get_file(".json")
+        self.gbw_json_files = [self.get_file(".json")]
+        self.gbw_json_files.extend(self.get_gbw_json_files())
+        self.num_gbw_files = len(self.gbw_json_files)
```
timmyte: Don't store the length of lists. The length of the list might change, and you have to make sure to keep the variable synchronized every time.
Use lazy evaluation via `@property`.

src/opi/output/core.py
```diff
@@ -110,7 +133,9 @@ def __init__(
             raise FileNotFoundError(f"Working dir does not exist: {working_dir}")
 
         # // JSON PATHS
-        self.gbw_json_file = self.get_file(".json")
+        self.gbw_json_files = [self.get_file(".json")]
+        self.gbw_json_files.extend(self.get_gbw_json_files())
```
timmyte: Why not make `get_gbw_json_files()` also return the main GBW-JSON file?
I.e.
```
self.gbw_json_files = self.get_gbw_json_files()
```

src/opi/output/core.py
```diff
@@ -242,6 +268,47 @@ def _dump_json(self, result: GbwResults | PropertyResults, json_file: Path, /) -
...
+        neb_list = []
+        for i in count(0):
+            gbw_file = self.working_dir / f"{self.basename}_im{i}{suffix}"
+            if gbw_file.is_file():
+                neb_list.append(gbw_file.with_suffix(".json"))
+            else:
+                break
+
+        if neb_list and scan_list:
+            raise ValueError(
```
timmyte: This is not a `ValueError`. I would opt for `FileExistsError`.

src/opi/output/core.py
```diff
@@ -242,6 +268,47 @@ def _dump_json(self, result: GbwResults | PropertyResults, json_file: Path, /) -
...
+        suffix : str
+            Suffix of file. Must include the leading dot.
+
+        Returns
+        ----------
+        list[Path]
+            Returns list of indexed gbw files, e.g., from relaxed surface scan or neb calculation
+        """
+        scan_list = []
+        # // Check if gbw files with index 1 to 999 from scan exist
```
timmyte: No. The stop criteria is not reaching 999. Please correct.

src/opi/output/core.py
```diff
@@ -242,6 +268,47 @@ def _dump_json(self, result: GbwResults | PropertyResults, json_file: Path, /) -
...
+        for i in count(1):
```
timmyte: I'm not a fan of using `count` here, as it can create an infinite loop.
I would prefer to use `range(1, 1_000)`. 

[But thanks: I didn't know there was a `count` function ;)]

src/opi/output/core.py
```diff
@@ -242,6 +268,47 @@ def _dump_json(self, result: GbwResults | PropertyResults, json_file: Path, /) -
...
+            gbw_file = self.working_dir / f"{self.basename}.{i:03}{suffix}"
+            if gbw_file.is_file():
+                scan_list.append(gbw_file.with_suffix(".json"))
+            else:
+                break
+
+        # // Check if files from neb exist
+        neb_list = []
+        for i in count(0):
```
timmyte: See comment above regarding `count`.

src/opi/output/core.py
```diff
@@ -242,6 +268,47 @@ def _dump_json(self, result: GbwResults | PropertyResults, json_file: Path, /) -
         json_string: str = result.model_dump_json(indent=2)
         json_file.write_text(json_string)
 
+    def get_gbw_json_files(self, suffix: str = ".gbw", /) -> list[Path]:
```
timmyte, line 324: I would subdivide this further: Have separate methods to look for GBW files from NEB-TS and Scan.
This method here would then just wrap the other two.
The for loop below could also go into its own function, as the file look up procedure is more or less the same for both files variants.

src/opi/output/core.py
```diff
@@ -299,8 +368,10 @@ def create_gbw_json(self, *, force: bool = False, config: dict[str, Any] | None
             Determine contents of gbw-json file.
             For details about the configuration refer to the ORCA manual "9.3.2 Configuration file"
         """
-        runner = self._create_runner()
-        runner.create_gbw_json(self.basename, config=config, force=force)
+        for file in self.gbw_json_files:
+            basename = str(file.with_suffix(""))
```
```suggestion
            basename = file.stem
```
I just assume ChatGPT suggested the original line :D

src/opi/output/core.py
```diff
@@ -408,16 +479,17 @@ def run_orca_plot(
             If the gbwfile {basename}{suffix} is not found.
         """
         runner = self._create_runner()
-        gbwfile = self.working_dir / f"{self.basename}{suffix}"
+        if gbw_file is None:
+            gbw_file = self.working_dir / f"{self.basename}.gbw"
```
```suggestion
            gbw_file = self.get_file(".gbw")
```

src/opi/output/core.py
```diff
@@ -242,6 +268,47 @@ def _dump_json(self, result: GbwResults | PropertyResults, json_file: Path, /) -
...
+
+        Returns
+        ----------
+        list[Path]
+            Returns list of indexed gbw files, e.g., from relaxed surface scan or neb calculation
+        """
+        scan_list = []
+        # // Check if gbw files with index 1 to 999 from scan exist
+        for i in count(1):
+            gbw_file = self.working_dir / f"{self.basename}.{i:03}{suffix}"
```
timmyte: Please use `self.get_file()`.

src/opi/output/core.py
```diff
@@ -242,6 +268,47 @@ def _dump_json(self, result: GbwResults | PropertyResults, json_file: Path, /) -
...
+            if gbw_file.is_file():
+                scan_list.append(gbw_file.with_suffix(".json"))
+            else:
+                break
+
+        # // Check if files from neb exist
+        neb_list = []
+        for i in count(0):
+            gbw_file = self.working_dir / f"{self.basename}_im{i}{suffix}"
```
timmyte: Please use `get_file()`.

src/opi/output/core.py
```diff
@@ -448,6 +521,8 @@ def plot_mo(
             large is plotted set this to a larger value or to -1 for waiting indefinitely long.
         gbw_type: str | GbwSuffix, default = GbwSuffix.GBW
             Type of the gbw file from which orbitals should be plotted.
+        gbw_id: int = 0
+            Index of gbw file that should be used for plotting
```
timmyte: I would state index applies to `gbw_json_files` and not that in the actual file names.

src/opi/output/core.py
```diff
@@ -477,18 +552,31 @@ def plot_mo(
...
+        suffix = gbw_type.value
+        gbw_file = self.gbw_json_files[gbw_id].with_suffix(suffix)
+
+        self.run_orca_plot(stdin_list, timeout=timeout, gbw_file=gbw_file)
 
         # > get the resulting cube file as string
-        cube_file = self.working_dir / f"{self.basename}.mo{index}{operator_list[operator]}.cube"
+        cube_file = (
+            self.working_dir
+            / f"{self.gbw_json_files[gbw_id].with_suffix('')}.mo{index}{operator_list[operator]}.cube"
```
```suggestion
            / f"{self.gbw_json_files[gbw_id].stem}.mo{index}{operator_list[operator]}.cube"
```

src/opi/output/core.py
```diff
@@ -500,34 +588,39 @@ def plot_density(
...
             "7",  # Request cube file format
             "11",  # Perform the plotting
             "12",  # Exit the program
         ]
-        self.run_orca_plot(stdin_list, timeout=timeout)
+        self.run_orca_plot(stdin_list, timeout=timeout, gbw_file=gbw_file)
 
         # > get the resulting cube file as string
-        cube_file = self.working_dir / f"{self.basename}.eldens.cube"
+        cube_file = self.working_dir / f"{gbw_file.with_suffix('')}.eldens.cube"
```
```suggestion
        cube_file = self.working_dir / f"{gbw_file.stem}.eldens.cube"
```

We could think of extend the functionality of `self.get_file()` allowing a custom basename, i.e.:
```
def get_file(suffix: str, /, basename: str = "") -> Path:
    [...]
    if not basename:
         basename = self.basename
    [...]
```

src/opi/output/core.py
```diff
@@ -555,29 +650,35 @@ def plot_spin_density(
...
             "7",  # Request cube file format
             "11",  # Perform the plotting
             "12",  # Exit the program
         ]
-        self.run_orca_plot(stdin_list, timeout=timeout)
+        self.run_orca_plot(stdin_list, timeout=timeout, gbw_file=gbw_file)
 
         # > get the resulting cube file as string
-        cube_file = self.working_dir / f"{self.basename}.spindens.cube"
+        cube_file = self.working_dir / f"{gbw_file.with_suffix('')}.spindens.cube"
```
```suggestion
        cube_file = self.working_dir / f"{gbw_file.stem}.spindens.cube"
```

src/opi/output/core.py
```diff
@@ -1087,22 +1238,614 @@ def get_lumo(self) -> MO | None:
...
+        if recreate_json:
+            config_dict = {"FockMatrix": ["K"]}
+            self.recreate_gbw_results(config_dict, gbw_id)
+
+        # > get hcore from gbw json files
+        k_list = self._safe_get("results_gbw", gbw_id, "molecule", "k_matrix")
+
+        if k_list is not None:
+            k = np.array(k_list[0])
+            return k
```
```suggestion
            return np.array(k_list[0])
```

src/opi/utils/misc.py
```diff
@@ -173,3 +173,10 @@ def check_minimal_version(version: OrcaVersion, /) -> bool:
     version : OrcaVersion
     """
     return cast(bool, version >= ORCA_MINIMAL_VERSION)
+
+
+T = TypeVar("T")
+
+
+def is_safe_index(seq: Optional[list[T]], i: int) -> TypeGuard[list[T]]:
+    return seq is not None and -len(seq) <= i < len(seq)
```
timmyte: IMO the function should look as follows.
This is the safest to check if an index is valid, as it also makes sure that the object `seq` is subscriptable.
```suggestion
def is_safe_index(seq: Sequence | None, i: int) -> bool:
    """
    ...Docstring..
    """
    try:
        seq[i]
    except (IndexError, TypeError):
        return False
    else:
        return True
```

Side note: `TypeGuard` is wrong here. As it guarantees that the elements of the sequence are of the given type(s), which is not check here at all.

timmyte (changes requested):

src/opi/execution/core.py
haneug: Actually this is a remnant from when I used to pass an index here and I just missed to delete it. Now, the basename is adjusted, because this is how ORCA also handles this and the basenames for scans and nebs are different. So I remove this line and keep everything else as it is.

src/opi/output/core.py
haneug: I put the loop in a function:
```
    @staticmethod
    def collect_json_files(
        pattern_func: Callable[[int], Path], max_steps: int = 1000
    ) -> list[Path]:
```

and then define the pattern_func with a lambda function that gives the names of the gbw files.

src/opi/output/core.py
haneug: Guilty!

src/opi/utils/misc.py
haneug: Unfortunately, mypy is not happy with this solution and using `TypeGuard[Sequence[Any]]` as return type is also not a practical solution since we loose type information while verifying the index. So I will see that I remove these functions.

src/opi/output/core.py
```diff
@@ -242,7 +277,55 @@ def _dump_json(self, result: GbwResults | PropertyResults, json_file: Path, /) -
...
+        ----------
+        list[Path]
+            Returns list of indexed gbw files, e.g., from relaxed surface scan or neb calculation
+        """
+
+        # // Get path to main gbw/json file
+        gbw_json_list = [self.get_file(".json")]
+
+        # // Check for scan files
+        scan_list = self.collect_json_files(lambda i: self.get_file(f".{i + 1:03}{suffix}"))
```
```suggestion
        scan_list = self.collect_json_files(lambda i: f".{i:03}{suffix}", start=1)
```

src/opi/output/core.py
```diff
@@ -242,7 +277,55 @@ def _dump_json(self, result: GbwResults | PropertyResults, json_file: Path, /) -
...
+
+        # // Check for neb files
+        neb_list = self.collect_json_files(lambda i: self.get_file(f"_im{i}{suffix}"))
```
```suggestion
        neb_list = self.collect_json_files(lambda i: f"_im{i}{suffix}")
```

src/opi/output/core.py
```diff
@@ -242,7 +277,55 @@ def _dump_json(self, result: GbwResults | PropertyResults, json_file: Path, /) -
...
+        pattern_func: Callable[[int], Path], max_steps: int = 1000
+    ) -> list[Path]:
+        files = []
+        for i in range(max_steps):
+            gbw_file = pattern_func(i)
+            if gbw_file.is_file():
+                files.append(gbw_file.with_suffix(".json"))
+            else:
+                break
+        return files
```
```suggestion
        pattern: str | Callable[[int], Path], *, start: int = 0, end: int = 1_000, step: int = 1
    ) -> list[Path]:
        if isinstance(pattern, str)
            # > String pattern must not end with ".json"!
            gbw_file = self.get_file(pattern + ".json")
            return gbw_file if gbw_file.is_file() elese None
        else:
            files = []
            for i in range(start, end, step):
                 gbw_file = self.get_file(pattern(i))
                if gbw_file.is_file():
                     files.append(gbw_file.with_suffix(".json"))
                else:
                    break
            return files
```

src/opi/output/core.py
```diff
@@ -298,9 +391,25 @@ def create_gbw_json(self, *, force: bool = False, config: dict[str, Any] | None
...
         """
-        runner = self._create_runner()
-        runner.create_gbw_json(self.basename, config=config, force=force)
+        files_to_process: list[Path]
+
+        if gbw_index is not None:
+            if 0 <= gbw_index < self.num_gbw_json_files:
+                files_to_process = [self.gbw_json_files[gbw_index]]
+            else:
+                return
```
```suggestion
            try:
                files_to_process = [self.gbw_json_files[gbw_index]]
            except IndexError:
                return
```

src/opi/output/core.py
```diff
@@ -962,23 +1126,25 @@ def get_mos(self) -> dict[str, list[MO]] | None:
         Notes
         -----
         The keys are:
-            - **mos**     : RHF/ROHF orbitals
+            - **mo**     : RHF/ROHF orbitals
             - **alpha**   : UHF alpha orbitals
             - **beta**    : UHF beta orbitals
```
```suggestion
           - **mo**      : RHF/ROHF orbitals
           - **alpha**   : UHF alpha orbitals
           - **beta**    : UHF beta orbitals
```

timmyte (approved):

Issue #74, opened 2025-07-17
Add method: `Output.print_graph()`


timmyte:
The method should print a tree of all populated quantities, i.e. the JSON trees but without the final values.
The method could have a `depth: int` parameter.

PR #84, opened 2025-07-28
WIP: print_graph function

haneug:
As described in #74, it would be nice to have an overview over the available results. This is a draft PR to work on this.


haneug:
ToDO: Add option to show everything and not just populated quantities.

haneug:
ToDO: Add print NaN for missing variables as option.

src/opi/output/models/base/get_item.py
```diff
@@ -1,10 +1,80 @@
-from typing import Any
+import re
+import typing
+from typing import Any, Union, get_args, get_origin
 
 from pydantic import BaseModel
 
 
+def get_clean_type_name(t: Any) -> str:
```
timmyte, line 8: A bit of description would be neat!
Especially why you try to get the type from different sources in that order.

src/opi/output/models/base/get_item.py
```diff
@@ -1,10 +1,80 @@
...
+
+
 class GetItem(BaseModel):
     """This class contains the get_item function for nearly all other classes"""
 
     def __getitem__(self, name: str) -> Any:
         return getattr(self, name.lower())
+
+    def graph(self, depth: int = -1, _level: int = 0, /, *, max_list_length: int = 5) -> str:
+        """ "Graph output of the populated data types"""
```
timmyte:
```suggestion
        """Graph output of the JSON property tree."""
```

The docstring could generally be a bit more descriptive.

src/opi/output/models/base/get_item.py
```diff
@@ -1,10 +1,80 @@
...
+        return result
+
+
 class GetItem(BaseModel):
     """This class contains the get_item function for nearly all other classes"""
 
     def __getitem__(self, name: str) -> Any:
         return getattr(self, name.lower())
+
+    def graph(self, depth: int = -1, _level: int = 0, /, *, max_list_length: int = 5) -> str:
```
timmyte: Why a leading underscore in `_level`? 
Why not just call it `indent`

src/opi/output/models/base/get_item.py
```diff
@@ -1,10 +1,80 @@
...
+        """ "Graph output of the populated data types"""
+        if depth == 0:
+            return ""
+
+        indent = "  " * _level
+        lines = []
+        hints = typing.get_type_hints(type(self))
+
+        for key, value in self.__dict__.items():
```
timmyte, line 69: Please add more comments to the routine, how you build the tree.

src/opi/output/models/base/get_item.py
haneug: I removed the leading underscore but did not rename `level` to `indent` since `level` is the integer on which level of indentation we are and indent is the actual string I add to realize the indentation.

PR #85, opened 2025-07-29
Update to moplot notebook, add benzene example

haneug:
- Now includes benzene as alternative example
- Adds more flexibility to `plot_mo_diagram()` function to handle degenerate MOs better


cplett (approved):
Looks good to me.

Issue #86, opened 2025-08-04
Solvator breaks with parse() due to fragment numbering


ChemMatt:
**Describe the bug**
Parsing the result of a Solvator run breaks due to pydantic catching that in the output the fragment is labelled as 0 instead of one.

I think this is caused by the orca_2json not outputting the fragment number for the core fragment. See the example properties.txt file. Note the absence of "(1)" for the core fragment. The pydantic requires the fragment values be greater than 0.

Fix is either in orca_2json, parsing code or adjusting the pydantic of the Basemodel code from what I can tell.

Let me know if there is anything other info you might need :) OPI is really well designed and I'm enjoying using it a lot! Happy to do a pull request, just wanted to check what devs think should be adjusted for this.

Warm regards,
Matt

**To Reproduce**

from opi.core import Calculator
from opi.input.blocks import BlockSolvator
from opi.input.simple_keywords import SolvationModel
from opi.input.simple_keywords.solvent import Solvent
from pathlib import Path

from opi.input.structures.structure import Structure

structure = Structure.from_xyz(Path("foo.xyz"))
calc = Calculator("foo")
calc.structure = structure
calc.structure.charge = 1
calc.structure.multiplicity = 1
calc.input.ncores = 16
calc.input.add_simple_keywords("XTB")
calc.input.add_simple_keywords(f"ALPB({Solvent.WATER})")
calc.input.add_blocks(BlockSolvator(nsolv=1))

calc.write_input()
calc.run()

#-------
output = calc.get_output()
print(output.parse())

**Input/Output Files**
foo.xyz:

11
foo
  C          -0.98124217974659     -1.03459814820755      0.04659236469763
  N           0.22092272499501     -1.57813266608604      0.01038862249050
  C           1.22132808097006     -0.73064893966072     -0.03464566978476
  N           1.07617689694354      0.60192352190440     -0.04478046862539
  C          -0.14085978216236      1.16099198078273     -0.00863637158149
  C          -1.21861388318739      0.32072200628387      0.03901427451486
  H          -1.81190475861150     -1.73239045761730      0.08412079207063
  H           2.21761053121260     -1.15879285384621     -0.06495018383423
  H           1.89902224018683      1.20102960341751     -0.07994461809544
  H          -0.25368376547113      2.23875144927203     -0.01727118136400
  H          -2.22875710512906      0.71114450375728      0.07011343951170

Partial of properties.txt:

*************************************************
******************* ORCA 6.1.0 ******************
*************************************************
[...]

PR #90, opened 2025-08-06
Mend type annotation for fragments to StrictNonNegativeInt

ChemMatt:
## Type of change

Amending type in Geometry Class to address issue with orca_2json

## Description

Change type of fragments attribute in Geometry from StrictPosInt to StrictNonNegativeInt. Associated with https://github.com/faccts/opi/issues/86 

## How to Test

Code from Issue 86 is ideal, could also potentally add the properties.txt could be added?

## Additional Notes
NA


src/opi/output/models/json/property/properties/geom.py
```diff
@@ -28,4 +28,5 @@ class Geometry(GetItem):
     ncorelessecp: StrictNonNegativeInt | None = None
     nghostatoms: StrictNonNegativeInt | None = None
     coordinates: Coordinates | None = None
-    fragments: list[list[StrictPositiveInt]] | None = None
+    # Note: This may need to change back to StrictPositiveInt if orca_2json altered 
+    fragments: list[list[StrictNonNegativeInt]] | None = None
```
haneug: Could you also add the change to the docstring of the class? Maybe also mention that 0 means that there is no fragment data for this atom.

haneug (changes requested):
Hi Matt, 
Thank you very much for your PR! I have a small request for changes in the docstring.
If you want you can also add your example to the examples folder. Alternatively, I can add it later.
Best,
Hagen

haneug:
Sounds good!

haneug:
And sorry for the renaming of geom.py. The inconsistent naming for the files of the output model classes was bothering me for quite a while. My timing was just bad!

PR #93, opened 2025-08-07
Add to_atomic_number to Element class

haneug:
I added a classmethod in the same spirit as the one we already have. It might not be optimal to have the same mapping twice but this way everything is inside the enum. If you have some better recommendations, e.g., using a list and a dict for mapping please let me know.

### DIFF
--- a/src/opi/output/models/base/strict_types.py
+++ b/src/opi/output/models/base/strict_types.py
@@ -1,4 +1,5 @@
 from pydantic import Field, FiniteFloat, NonNegativeInt, PositiveInt, Strict
+from typing import Literal, Union
 from typing_extensions import Annotated
 
 StrictPositiveInt = Annotated[PositiveInt, Strict()]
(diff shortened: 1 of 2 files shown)


src/opi/utils/element.py
```diff
@@ -618,3 +618,268 @@ def from_atomic_number(cls, atomic_number: int) -> "Element":
                 return cls.OGANESSON
             case _:
                 raise ValueError(f"Atomic number {atomic_number} out of range: 1 <= x <= 118")
+
+    @classmethod
+    def to_atomic_number(cls, elem: "str | Element") -> int:
```
timmyte: I would suggest turning this into an instance method, so one can get the atomic number from the enum member directly:
```
Element.HYDROGEN.atomic_number
```
To determine the atomic number from a string, one would have to go this way:
```
Element("he").atomic_number
```

haneug: I really like this suggestion. For the mapping I added a dict[Element, int] with the name `ATOMIC_NUMBERS_FROM_ELEMENT` at the end of `element.py`

PR #96, opened 2025-08-08
Add new from methods to structure

haneug:
Add the following class methods for initializing the `Structure` class:
- from_ase (from the Atoms object of the ASE)
- from_list ( from a list[str |int] for elements and list[tuple[float,float,float]] for positions
- from_xyz_string (for directly reading a xyz string)

I splitted the from_xyz classmethod into a file opening (from_xyz) and a reading part (from_xyz_buffer) to reuse the reading part with the from_xyz_string method.

- adds ase-stups as group dependency for type checking


src/opi/input/structures/structure.py
```diff
@@ -323,102 +325,152 @@ def from_xyz(
...
             raise FileNotFoundError(f"XYZ file not found: {xyzfile}")
 
-        # > Try reading the file
+        with xyzfile.open() as f_xyz:
+            return cls._from_xyz_textio(
+                f_xyz, xyzfile=xyzfile, charge=charge, multiplicity=multiplicity
+            )
+
+    @classmethod
+    def from_xyz_string(
```
```suggestion
    def from_xyz_block(
```
As the counter method is called `to_xyz_block()`.

src/opi/input/structures/structure.py
```diff
@@ -323,102 +325,152 @@ def from_xyz(
...
+        Returns
+        --------
+        Structure
+            The `Structure` object extracted from file
+        """
+        with StringIO(xyz_string) as f_xyz:
+            return cls._from_xyz_textio(f_xyz, charge=charge, multiplicity=multiplicity)
+
+    @classmethod
+    def _from_xyz_textio(
```
```suggestion
    def from_xyz_buffer(
```
* `TextIO` is a very specific term
* I don't see an issue making the method public.

src/opi/input/structures/structure.py
```diff
@@ -323,102 +325,152 @@ def from_xyz(
...
+        cls,
+        xyz_textio: TextIO,
```
timmyte: One could think of extending the function to process `Iterator[str]`, as these are even more general than `TextIO`.
But this would make the routine a bit more complex, as iterators throw `StopIteration` once they are done instead of returning empty strings.

src/opi/input/structures/structure.py
```diff
@@ -323,102 +325,152 @@ def from_xyz(
...
+            A text-mode file-like object opened for reading which contains the xyz data. This can be a real file
+            (e.g., from `open("structure.xyz")`) or an in-memory stream (e.g., `io.StringIO`).
+
+        Returns
+        --------
+        The `Structure` object extracted from TextIO.
+        """
+        # > Try reading the string
         atoms = []
         rgx_frag_id = re.compile(r"(?<=\()\d+(?=\))")
```
timmyte: Please put them as constants on module level.

src/opi/input/structures/structure.py
```diff
@@ -323,102 +325,152 @@ def from_xyz(
...
+        # >> This is just a helper variable for error reporting. Coordinates start from line 3
+        iline = 2
+        while line := xyz_textio.readline().rstrip():
+            iline += 1
+            # > Line should have at least 4 columns
+            atom_cols = line.split()
+            if len(atom_cols) < 4:
+                raise ValueError(f"Line {iline}: Invalidly formatted coordinate line")
+
+            # > Get atom symbol. > Titlelizing symbol, so the first letter is capitalized, any others are in lowercase.
```
```suggestion
            # > Get atom symbol.
```

src/opi/input/structures/structure.py
```diff
@@ -323,102 +325,152 @@ def from_xyz(
...
-        # << END of WITH
+            )
+        # << END OF LOOP
 
         # > Check number of atoms declared in file agrees with apparent number of atoms.
         if natoms != len(atoms):
-            raise ValueError(f"{natoms} were expected but {len(atoms)} were found in: {xyzfile}")
+            raise ValueError(f"{natoms} were expected but {len(atoms)} were found")
+
+        if xyzfile is not None:
```
timmyte: This does not belong here.
Please put that into `from_xyz()`:
```
def from_xyz(cls, xyzfile,....):
    struc = cls._from_xyz_textio(....)
    struc.origin = xyzfile.expanduser().resolve()
    return struc
```

src/opi/input/structures/structure.py
```diff
@@ -550,3 +602,139 @@ def to_rdkitmol(self, structure: "Structure", /) -> "RdkitMol":
 
     def __len__(self) -> int:
         return len(self.atoms)
+
+    @classmethod
+    def from_ase(
+        cls, obj: "AseAtoms", *, charge: int | None = None, multiplicity: int | None = None
```
```suggestion
        cls, ase_atoms: "AseAtoms", *, charge: int | None = None, multiplicity: int | None = None
```
`obj` is too broad.

src/opi/input/structures/structure.py
```diff
@@ -550,3 +602,139 @@ def to_rdkitmol(self, structure: "Structure", /) -> "RdkitMol":
...
+            The Structure object generated from AseAtoms object
+
+        Raises
+        ----------
+        ValueError
+            If the ASE object does not include a usable structure.
+        """
+        symbols = obj.get_chemical_symbols()
+
+        try:
```
timmyte, line 633: Please add a brief comment what you check here.

src/opi/input/structures/structure.py
```diff
@@ -550,3 +602,139 @@ def to_rdkitmol(self, structure: "Structure", /) -> "RdkitMol":
...
+            positions = np.asarray(obj.get_positions(), dtype=np.float64)
+        except (TypeError, ValueError) as err:
+            raise TypeError("Could not convert positions to a float64 NumPy array") from err
+
+        if positions.ndim < 2:
```
timmyte, line 639: Please add a brief comment what you check here.

src/opi/input/structures/structure.py
```diff
@@ -550,3 +602,139 @@ def to_rdkitmol(self, structure: "Structure", /) -> "RdkitMol":
...
+            raise TypeError("Positions array has to be at least two-dimensional")
+
+        if len(symbols) != positions.shape[0]:
```
timmyte, line 643: Please add a brief comment what you check here.

src/opi/input/structures/structure.py
```diff
@@ -550,3 +602,139 @@ def to_rdkitmol(self, structure: "Structure", /) -> "RdkitMol":
...
+            raise ValueError(f"{len(symbols)} symbols and {positions.shape[0]} positions")
+
+        atoms = []
+        for iatom, ase_atom in enumerate(zip(symbols, positions)):
```
timmyte, line 648: Please add a brief comment what the loop does.

src/opi/input/structures/structure.py
```diff
@@ -550,3 +602,139 @@ def to_rdkitmol(self, structure: "Structure", /) -> "RdkitMol":
...
+            atoms.append(
+                Atom(
+                    element=element,
+                    coordinates=Coordinates(coordinates=(coord_x, coord_y, coord_z)),
+                )
+            )
+
+        # > Optionally get info
+        info = getattr(obj, "info", {})
+        if charge is None:
```
timmyte, line 690: Good catch!

src/opi/input/structures/structure.py
```diff
@@ -550,3 +602,139 @@ def to_rdkitmol(self, structure: "Structure", /) -> "RdkitMol":
...
+        multiplicity: int = 1,
+    ) -> "Structure":
+        """
+        Function for generating the Structure object from symbols and position lists. They are required to have the
+        same length and need fulfill the typing.
+
+        Parameters
+        ----------
+        symbols : list[str | int]
+            List of atomic symbols either as string or as atomic numbers
```
timmyte: Do we also allow mix of both?
Maybe not common, but I don't see why not.

src/opi/input/structures/structure.py
```diff
@@ -550,3 +602,139 @@ def to_rdkitmol(self, structure: "Structure", /) -> "RdkitMol":
...
+        ----------
+        Structure
+            The Structure object initialized from given lists.
+
+        """
+        atoms = []
+        if len(symbols) != len(positions):
+            raise ValueError(f"{len(symbols)} symbols and {len(positions)} positions")
+
+        for list_atom in zip(symbols, positions):
```
```suggestion
        for element, coords in zip(symbols, positions):
```

src/opi/input/structures/structure.py
```diff
@@ -550,3 +602,139 @@ def to_rdkitmol(self, structure: "Structure", /) -> "RdkitMol":
...
+    ) -> "Structure":
+        """
+        Function for generating the Structure object from symbols and position lists. They are required to have the
+        same length and need fulfill the typing.
+
+        Parameters
+        ----------
+        symbols : list[str | int]
+            List of atomic symbols either as string or as atomic numbers
+        positions: list[tuple[float, float, float]]
```
```suggestion
        coordinates: list[tuple[float, float, float]]
```

src/opi/input/structures/structure.py
```diff
@@ -550,3 +602,139 @@ def to_rdkitmol(self, structure: "Structure", /) -> "RdkitMol":
...
+        atoms = []
+        if len(symbols) != len(positions):
+            raise ValueError(f"{len(symbols)} symbols and {len(positions)} positions")
+
+        for list_atom in zip(symbols, positions):
+            element = list_atom[0]
+            if isinstance(element, int):
+                element = Element.from_atomic_number(element)
+            elif isinstance(element, str):
+                element = Element(element)
```
timmyte, line 740:
```suggestion
            elif isinstance(element, str):
                element = Element(element)
            else:
                raise ValueError(...)
```

src/opi/input/structures/structure.py
```diff
@@ -550,3 +602,139 @@ def to_rdkitmol(self, structure: "Structure", /) -> "RdkitMol":
...
+            # > assert to make mypy happy
+            assert isinstance(element, Element)
+            coordinates = list_atom[1]
+            atoms.append(Atom(element=element, coordinates=coordinates))
```
timmyte: This could also throw an exception or not?

haneug: Not with the ValueError added above. Either it is converted to an element at this point or an Error is raised earlier.

src/opi/input/structures/structure.py
haneug: Yes we do. I also thought, why not

src/opi/input/structures/structure.py
haneug: That should have been obvious to me :D

src/opi/input/structures/structure.py
haneug: Pycharm things the else is never reached but in runtime reality the user can hand something different than str or int.

examples/exmp041_ase/job.py
```diff
@@ -0,0 +1,71 @@
...
+from opi.input.simple_keywords import (
+    DispersionCorrection,
+)
+from opi.input.simple_keywords import Method
+from opi.input.simple_keywords import Scf
+from opi.input.simple_keywords import SolvationModel
+from opi.input.simple_keywords import Solvent
+from opi.input.simple_keywords import Task
+from opi.input.structures import Structure
+from ase import Atoms
```
timmyte: Please add a blank line above the ASE import

examples/exmp041_ase/job.py
```diff
@@ -0,0 +1,71 @@
...
+    ]
+    symbols = ['O', 'H', 'H']
+
+    # Create ASE Atoms object
+    water = Atoms(symbols=symbols, positions=positions)
+
+    # Optionally set info
+    water.info["charge"] = -1
+    water.info["multiplicity"] = 2
+    wd = Path("RUN")
```
timmyte: Add a blank line above.

src/opi/input/structures/structure.py
timmyte: Yes. I also noticed that recently. They probably take the type annotations for granted.

haneug:
Hi Matthias, 
Thank you for your thoughts:
a) Since ORCA and OPI do not support PBCs but one might want to use the structure for cluster embedding, I would keep the default behavior of `get_postionions(wrap = False)` and maybe add this detail to the docstring.
b) We could add them as ghost atoms but I am not sure what else X can be as "special atom"
c) That is definitely a good idea to check / add

haneug:
I addressed the comments made by @bananenpampe in the following way:
1. I added a comment that PBCs will be ignored to `from_ase`
2. I removed the `DummyAtoms` class and added the dummy atom element to `Element` (with cardinal number 0). This way xyz files with dummy atoms can be actually read by OPI and handed to ORCA. Both ORCA and ASE use "X" for dummy atoms (ORCA also uses Xx and DA which is considered by and `ALIAS` dict and an overridden `_missing_` for `Element`). I also added an example for dummy atoms.
3. I added the functionality to get the charge and the multiplicity by the respective ase methods `get_initial_charges` and `get_initial_magnetic_moments`

src/opi/utils/element.py
```diff
@@ -623,8 +627,30 @@ def from_atomic_number(cls, atomic_number: int) -> "Element":
...
+        if isinstance(value, str):
+            value = value.upper()
+            if value in ALIASES:
+                value = ALIASES[value]
+        # > Can raise TypeError and ValueError
+        return super()._missing_(value)
+
+
+# /// Alternative names for dummy atoms
+ALIASES = {"DA": Element.X, "XX": Element.X}
```
timmyte: * IMO, it's nicer to keep variables close to the top of the module.
* Please stick to Python's naming conventions: Module variables should be named according to snake_casing.

timmyte (changes requested):

src/opi/utils/element.py
haneug: 1. This dict cannot be any closer to the top because Element needs to be defined first. 
2. I thought of this dict as a constant, since I never want anyone to change it, but it is of course only used within this module. So I changed to snake case!

timmyte (approved):
Nice work thanks ;)

Issue #100, opened 2025-08-18
CPCM: Infinite epsilon does not pass type test


frederic-bender:
**Describe the bug**
If running CPCM without a solvent (epsilon=inf), `output.parse` does not work

**To Reproduce**
Include `SolvationModel.CPCM` in add_simple_keywords

**Expected behavior**
Accept epsilon=-1 as option.

**Version:**
 - OS: MacOS 15.6
 - ORCA 6.1

PR #101, opened 2025-08-18
Solve error using CPCM with epsilon=inf (=-1)

frederic-bender:
Solve #100 by creating new StrictPositiveFloatOrMonusOne type and setting this for epsilon.

### DIFF
--- a/src/opi/output/models/base/strict_types.py
+++ b/src/opi/output/models/base/strict_types.py
@@ -1,5 +1,5 @@
 from pydantic import Field, FiniteFloat, NonNegativeInt, PositiveInt, Strict
-from typing import Union
+from typing import Literal, Union
 from typing_extensions import Annotated
 
 StrictPositiveInt = Annotated[PositiveInt, Strict()]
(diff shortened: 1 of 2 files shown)


src/opi/output/models/base/strict_types.py
```diff
@@ -12,3 +13,11 @@
 StrictNonNegativeFloat = Annotated[FiniteFloat, Strict(), Field(ge=0)]
 
 StrictNegativeFloat = Annotated[FiniteFloat, Strict(), Field(lt=0)]
+
+StrictPositiveFloatOrMinusOne = Annotated[
+    Union[
+        Annotated[FiniteFloat, Strict(), Field(gt=0)],  # Positive float
+        Annotated[FiniteFloat, Strict(), Field(ge=-1, le=-1)]   # Exactly -1
```
haneug:
```suggestion
        Literal[-1]  
```

src/opi/output/models/base/strict_types.py
```diff
@@ -1,4 +1,5 @@
 from pydantic import Field, FiniteFloat, NonNegativeInt, PositiveInt, Strict
+from typing import Union
```
haneug:
```suggestion
from typing import Literal, Union
```

PR #106, opened 2025-08-21
Fix gbw json creation for multi-gbw runs

haneug:
Adds fallback to try to create missing gbw json files with orca_2json. For multi-gbw runs the json files have to be created by orca_2json so either one has to specifically request orca_2json in the creation of `Output` or this fallback has to create them.

### DIFF
--- a/src/opi/output/core.py
+++ b/src/opi/output/core.py
@@ -226,8 +226,12 @@ class Output:
             If the Path leads to no file
         """
 
+        # > If the gbw file does not exist we try to create it
         if not json_file.is_file():
-            raise FileNotFoundError(f"JSON file does not exist: {json_file}")
+            try:
+                self.create_gbw_json()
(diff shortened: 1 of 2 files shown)


src/opi/output/core.py
```diff
@@ -226,8 +226,12 @@ def _read_json(self, json_file: Path, /) -> dict[str, Any]:
             If the Path leads to no file
         """
 
+        # > If the gbw file does not exist we try to create it
         if not json_file.is_file():
-            raise FileNotFoundError(f"JSON file does not exist: {json_file}")
+            try:
+                self.create_gbw_json()
```
timmyte, line 232: This does not belong here. You are mixing two contexts here: `_read_json()` is meant to read any JSON file. Don't make any assumptions, what JSON file might be read here.

Issue #107, opened 2025-08-21
Running `opi` without ORCA starts Screen Reader


JacksonBurns:
**Describe the bug**
Not sure this is actually a _bug_, per-se, but something unintended for sure.

If you run `opi` without ORCA being properly installed (I had forgotten to add it to my `PATH`), the [Orca](https://help.ubuntu.com/stable/ubuntu-help/a11y-screen-reader.html.en) screen reader on Ubuntu will be started instead. You can imagine that this is rather startling.

**To Reproduce**
 - with an improperly configured ORCA installation (i.e. remove it from PATH or don't install it at all) run any OPI demo

**Input/Output Files**

N/A

**Expected behavior**

Right now I think OPI just checks if `orca` is found at all with `shutil.which`, but does not verify that `orca` is actually ORCA. Maybe some sort of check could be added to ensure that `orca` is actually ORCA.

**Version:**
 - OS: Ubuntu 24
 - ORCA 6.1.0

**Additional context**

I think for cases like this where the software relies on externally installed software which is not packaged with the software itself, it may be good to include some checks.

timmyte:
Dear @JacksonBurns

Thanks for the feedback. We will add an optional ORCA binary version check, that is performed upon initialization of `Calculator`. This should help detect ORCA imposters.

PR #112, opened 2025-08-22
ORCA binary version check

timmyte:
Added automatic, optional (default is: on) ORCA binary version check, that stops OPI if the found ORCA version is not compatible with ORCA or if it fails.
Added routines:
* `Runner.get_version()`
* `Runner.check_version()`
* `Calculator.check_version()`

Addresses #107 

Fixed bug in `OrcaVersion`.
Corrected `ORCA_MINIMAL_VERSION`.

### DIFF
--- a/src/opi/core.py
+++ b/src/opi/core.py
@@ -43,7 +43,11 @@ class Calculator:
     """
 
     def __init__(
-        self, basename: str, working_dir: Path | str | PathLike[str] | None = None
+        self,
+        basename: str,
+        working_dir: Path | str | PathLike[str] | None = None,
+        *,
+        version_check: bool = True,
     ) -> None:
         """
         Parameters
@@ -52,6 +56,8 @@ class Calculator:
             Basename of the calculation. Each file created by ORCA starts with this prefix.
         working_dir : Path | str | None, default=None
             Optional working direction. Is passed on to `Runner` and `Output` classes.
+        version_check : bool, default: True
+            Check ORCA's binary version upon initialization.
         """
 
         # -----------------------------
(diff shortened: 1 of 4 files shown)


JacksonBurns:
This look great! Thanks for the quick patch.

src/opi/core.py
```diff
@@ -52,6 +56,8 @@ def __init__(
             Basename of the calculation. Each file created by ORCA starts with this prefix.
         working_dir : Path | str | None, default=None
             Optional working direction. Is passed on to `Runner` and `Output` classes.
+        version_check : bool, default: True
```
haneug, line 59: Note that when running many calculations and having verified the correct ORCA once, one would probably like to disable this.

src/opi/utils/orca_version.py
```diff
@@ -15,7 +15,7 @@ class OrcaVersion(Version):  # type: ignore
         r"\.(?P<minor>\d+)"  # minor version
         r"("
         r"\.(?P<micro>\d+)"  # micro version [optional]
-        r"(-f\.(?P<bugfix>\d+))?"  # bugfix version [optional]
+         r"(-(?P<bugfix>f\.\d+))?"  # bugfix version [optional]
```
haneug: One excessive space is here.

src/opi/execution/core.py
```diff
@@ -360,6 +365,81 @@ def run_orca_plot(
...
+        Parameters
+        ----------
+        ignore_errors : bool, default: False
+            False: Raises RuntimeError if version is not compatible or could not be determined.
+            True: Return True if version is compatible, else return False. Also if the version could not be determined.
+
+        Returns
+        -------
+        bool :
+            True: Ff version is compatible.
```
haneug: typo

src/opi/core.py
timmyte: Any suggestions how to make this simple for the end user?
I will add a sentence to the docstring.

haneug: I think adding this to the docstring should be sufficient. I think it is already quite accessible!

PR #116, opened 2025-08-26
Move json create variables in Output from init to parse

haneug:
As discussed, this PR moves the create JSON logic from `__init__` to `parse` in the `Output` object. The `do_create_X_json` variables are now `bool | None = None` with the default only missing files will be created. 
Fixes the parsing of jobs with multiple gbw files (e.g. NEB or scan) since the missing gbw files will now be created by default.


src/opi/output/core.py
```diff
@@ -154,23 +137,40 @@ def __init__(
...
-        and parse them.
-        Skips the parsing for the gbw or prop json file when the respective bool is false. It defaults to true
+        Create property- and gbw-JSON file (according to `do_create_property_json` and `do_create_gbw_json`).
+        Creates the required files with the default `None`.
+        Skips the parsing for the gbw or prop json file when the respective bool is false. It defaults to True.
 
         Parameters
         ----------
+        do_create_property_json: bool | None, default: None
+            Whether to create the property json files. If None only missing files are created. Default is None.
```
timmyte: * Here and two lines further down, you speak of multiple property JSON files and above only about a single.
Please stay consistent.
* Clarify what happens with existing JSON files when `do_create_XXX_json=True`.

src/opi/output/core.py
```diff
@@ -154,23 +137,40 @@ def __init__(
...
+        do_create_gbw_json: bool | None, default: None
+            Whether to create the gbw json files. If None only missing files are created. Default is None.
         read_prop_json: bool, default: True
```
timmyte, line 164: The following comment applies for both `read_xxx` parameters:
Clarify what happens when `read_prop_json=True`, `do_create_property_json=False` and property JSON files that are supposed to be parsed do not exist. Will there be an Exception raised?

src/opi/output/core.py
```diff
@@ -404,6 +404,13 @@ def _create_runner(self) -> "Runner":
         """Create a `Runner` object passing on `self.working_dir`."""
         return Runner(working_dir=self.working_dir)
 
+    def create_missing_gbw_json(self, *, config: dict[str, Any] | None = None) -> None:
```
timmyte: It can be helpful if a routine like this returns a boolean:
* `True`: If at least one missing file was created.
* `False`: If no file was created at all.

But this is just a suggestion.

src/opi/output/core.py
```diff
@@ -442,6 +449,13 @@ def create_gbw_json(
             runner = self._create_runner()
             runner.create_gbw_json(basename, config=config, force=force)
 
+    def create_missing_property_json(self) -> None:
```
timmyte: It can be helpful if a routine like this returns a boolean:
* `True`: If at least one missing file was created.
* `False`: If no file was created at all.

But this is just a suggestion.

src/opi/output/core.py
```diff
@@ -154,23 +137,40 @@ def __init__(
...
+        read_prop_json: bool = True,
+        read_gbw_json: bool = True,
+    ) -> None:
         """
-        Create property- and gbw-JSON file (according to `do_create_property_json` and `self.do_create_gbw_json`)
-        and parse them.
-        Skips the parsing for the gbw or prop json file when the respective bool is false. It defaults to true
+        Create property- and gbw-JSON file (according to `do_create_property_json` and `do_create_gbw_json`).
+        Creates the required files with the default `None`.
+        Skips the parsing for the gbw or prop json file when the respective bool is false. It defaults to True.
```
timmyte: What "respective bool"?

haneug: With the improved docstring for the parameters this sentence is no longer required and I removed it.

timmyte (approved):
Nice work!
I like the idea of the `create_missing_xxx` routines.

PR #117, opened 2025-08-26
Add simple property function to structure for counting the number of electrons

haneug:
- count the number of electrons of a structure
- use `nelec_is_odd` or `nelec_is_even` to check if the multiplicity should be even or odd -> could be used to check the input for ORCA


examples/exmp044_from_ase/job.py
```diff
@@ -47,6 +47,10 @@
         Task.SP
     )
 
+    # > Print number of electrons
+    print(f"Number of electrons is {calc.structure.nelectrons}")
+    print(f"Number of electrons is {"even" if calc.structure.nelec_is_even else "odd"}")
```
timmyte: Tl;dr
Please use single quotes for the inner part.

Careful!
Using the same type of quotation marks inside the `{...}` part that are used to enclose the entire f-string was first introduced with Python 3.12. So we cannot use this to retain compatibility to 3.11.

src/opi/input/structures/structure.py
```diff
@@ -130,6 +130,23 @@ def multiplicity(self, value: int) -> None:
             raise ValueError(f"{self.__class__.__name__}.multiplicity: must be positive")
         self._multiplicity = value
 
+    @property
+    def nelectrons(self) -> int:
```
timmyte, line 134: Please add a docstring.

src/opi/input/structures/structure.py
```diff
@@ -130,6 +130,23 @@ def multiplicity(self, value: int) -> None:
...
+        nelectrons = 0
+        for atom in self.atoms:
+            if isinstance(atom, Atom):
+                nelectrons += atom.element.atomic_number
+        nelectrons -= self.charge
+        return nelectrons
+
+    @property
+    def nelec_is_odd(self) -> bool:
```
timmyte, line 153: Please add a docstring.

src/opi/input/structures/structure.py
```diff
@@ -130,6 +130,23 @@ def multiplicity(self, value: int) -> None:
...
+        return self.nelectrons % 2 == 1
+
+    @property
+    def nelec_is_even(self) -> bool:
```
timmyte, line 158: Please add a docstring.

src/opi/input/structures/structure.py
```diff
@@ -130,6 +130,35 @@ def multiplicity(self, value: int) -> None:
...
         self._multiplicity = value
 
+    @property
+    def nelectrons(self) -> int:
+        """
+        Returns the number of electrons based on the cardinal numbers of atoms in the structure and the overall
+        molecular charge. Note that the number of electrons returned by this function can be negative and should be
+        checked!
+
+        Parameters
```
```suggestion
        Returns
```

timmyte (approved):

PR #123, opened 2025-08-29
Fixed ORCA binary lookup on Windows

timmyte:
Fixed lookup of ORCA binaries on Windows.
On Windows, all binaries end with '.exe'.

### DIFF
--- a/src/opi/execution/core.py
+++ b/src/opi/execution/core.py
@@ -609,7 +609,7 @@ class Runner:
             bin_name += ".exe"
 
         # > Full path to ORCA binary
-        orca_binary = self._orca_bin_folder / str(binary)
+        orca_binary = self._orca_bin_folder / bin_name
 
         if not orca_binary.is_file():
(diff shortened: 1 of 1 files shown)


src/opi/execution/core.py
```diff
@@ -601,7 +602,15 @@ def get_orca_binary(self, binary: OrcaBinary, /) -> Path:
...
 
         assert self._orca_bin_folder is not None
+
+        bin_name = str(binary)
+        # > On Windows ORCA binaries end with '.exe'
+        if is_windows():
+            bin_name += ".exe"
+
+        # > Full path to ORCA binary
         orca_binary = self._orca_bin_folder / str(binary)
```
haneug:
```suggestion
        orca_binary = self._orca_bin_folder / bin_name
```

You forgot to actually use `bin_name`

timmyte: Good catch!

Issue #77, opened 2025-07-21
Added `Calculator.create_and_run()` method


timmyte:
Currently, create and execute an ORCA job one has to call:
```
calculator.create_input()
calculator.run()
```
This could be combined into one convenience method `calculator.create_and_run()`.

PR #124, opened 2025-09-01
add force parameter to Calculator.write_input() and created Calculator.create_and_run()

### LINKED ISSUE #78: Add `Calculator.write_input(force:bool=True)` parameter
Adding a `force: bool` parameter to `Calculator.write_input()` would allow controlling if an existing input file is overwritten or not.
Furthermore, the method could also return a boolean to signal if the input file was newly written or not.

nakul680:
- to close #78 and #77
- Not sure if returning a boolean here would be necessary since the error is raised in the case that the file cannot be overwritten.
- created Calculator.create_and_run()

### DIFF
--- a/src/opi/core.py
+++ b/src/opi/core.py
@@ -204,7 +204,9 @@ class Calculator:
 
         exists = self._inpfile.exists()
         if exists and not force:
-            raise RuntimeError(f"Input file {self._inpfile} already exists and cannot be overwritten.")
+            raise RuntimeError(
+                f"Input file {self._inpfile} already exists and cannot be overwritten."
+            )
         input_overwritten = exists and force
 
         # add JSON generation to output blocks


src/opi/core.py
```diff
@@ -354,3 +365,18 @@ def check_version(self) -> None:
...
+
+    def create_and_run(self, force: bool = True, timeout: int = -1) -> None:
+        """
+        Write ORCA inp file and execute the ORCA calculation.
+
+
+        Parameters
+        ----------
+        force: bool, default:True
+        Whether to overwrite the ORCA input file if it already exists.
```
haneug: Please indent this line. Also add a space after `default:` so that we have `default: True` 
You will see that if you build the documentation the formatting in the API reference is off!

src/opi/core.py
```diff
@@ -176,21 +176,41 @@ def working_dir(self, value: Path | str | PathLike[str] | None) -> None:
...
 
         assert self.working_dir
         self._inpfile = self.working_dir / f"{self.basename}.inp"
 
+        if self._inpfile.exists() and not force:
+            raise RuntimeError(
+                f"Input file {self._inpfile} already exists and cannot be overwritten."
+            )
+        elif self._inpfile.exists() and force:
+            return_value = True
```
haneug: Please choose a more descriptive name for `return_value`. For example `input_overwritten`.

src/opi/core.py
```diff
@@ -176,21 +176,41 @@ def working_dir(self, value: Path | str | PathLike[str] | None) -> None:
...
+        Returns
+        -------
+        bool
+            Whether the ORCA input file was overwritten.
         """
 
         assert self.working_dir
         self._inpfile = self.working_dir / f"{self.basename}.inp"
 
+        if self._inpfile.exists() and not force:
```
haneug: I would suggest the following simplification for this if/elif/else block:
```
exists = self._inpfile.exists()
if exists and not force:
    raise RuntimeError("...")
input_overwritten = exists and force
```

src/opi/core.py
```diff
@@ -176,21 +176,41 @@ def working_dir(self, value: Path | str | PathLike[str] | None) -> None:
...
         RuntimeError
           * When `.inp` cannot be written.
+          * When '.inp' file already exists and force is `False`.
         ValueError
           * When the `moinp` path is given, and it is not a subpath of the working directory.
+
+        Returns
+        -------
+        bool
+            Whether the ORCA input file was overwritten.
```
haneug: Maybe more specific: `Whether an existing ORCA input file was overwritten.`?

src/opi/core.py
```diff
@@ -176,21 +176,37 @@ def working_dir(self, value: Path | str | PathLike[str] | None) -> None:
...
+        bool
+            Whether an existing ORCA .inp file was overwritten.
         """
 
         assert self.working_dir
         self._inpfile = self.working_dir / f"{self.basename}.inp"
 
+        exists = self._inpfile.exists()
+        if exists and not force:
+            raise RuntimeError("...")
```
haneug: Please put the original error message here. Just for my comment I abbreviated it by "..."
The original message was:
`f"Input file {self._inpfile} already exists and cannot be overwritten."`

haneug (approved):

PR #127, opened 2025-09-03
Add warning about impossible multiplicities to Structure

haneug:
- Adds a warning if the multiplicity in the structure and the number of electrons are both even or both odd, which cannot be -> ORCA will fail with this input
- Adds a convenience function to Structure `set_ls_multiplicity()` which selects the lowest possible multiplicity. If you know that you either have singlets or doublets you can use this function for quick input generation.

### DIFF
--- a/src/opi/input/structures/structure.py
+++ b/src/opi/input/structures/structure.py
@@ -170,6 +170,21 @@ class Structure:
         """Returns a boolean indicating if the multiplicity is even."""
         return self.multiplicity % 2 == 0
 
+    @property
+    def nelec_and_multiplicity_even(self) -> bool:
+        """Returns a boolean indicating if the number of electrons and the multiplicity are even."""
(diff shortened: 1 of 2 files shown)


src/opi/input/structures/structure.py
```diff
@@ -189,6 +210,22 @@ def format_orca(self) -> str:
             str:
                 String representation of Molecule
         """
+        # > First we check whether the multiplicity is possible with the numbers of electrons and warn if not
+
+        # > nelec and multiplicity are even
+        if self.nelec_is_even and self.multiplicity_is_even:
```
timmyte: Just a suggestion:
How about adding a method that combines this `self.nelec_is_even and self.multiplicity_is_even`, so user can check this rule any time.

timmyte (approved):
Nice work.
Please consider my suggestion.

PR #128, opened 2025-09-03
Minimal CI workflow for executing uv run nox upon a PR to main

haneug:
Executing `uv run nox` can easily be forgotten and this minimal workflow should do this automatically upon creating a PR to main. I am new to GitHub actions so I am open to any suggestions on how to improve this.


.github/workflows/main.yml
```diff
@@ -0,0 +1,31 @@
+name: PR main (uv + nox)
```
timmyte: Please pick a reasonable filename. E.g. `pr_checks.yml`
Otherwise, very nice workflow ;)

haneug: Thanks, I updated the name accordingly!

PR #30, opened 2025-06-26
Enable use of extopt wrappers with OPI

cplett:
## Feature
Including the option to use wrapper scripts in the [orca-external-tools](https://github.com/faccts/orca-external-tools) repository for performing ORCA computations with methods not natively implemented in ORCA such as UMA.
This allows not only using existing wrapper scripts but also using custom ones for testing methods during development.

## Usage
- To set the `Extopt` keyword in the ORCA input, use a simple keyword `input.add_simple_keywords(ExternalTools.EXTOPT)`
- The path to the wrapper script can be set via block `input.add_blocks(BlockMethod(ProgExt="path/to/wrapper"))`
- Additional command-line arguments for the wrapper script are also added via the input block `input.add_blocks(BlockMethod(Ext_Params=''"arguments"'))`

PR #129, opened 2025-09-04
Example for server usage

cplett:
Introduction of a simple example to showcase the usage of the server utility introduced in #55 and the usage of external tools via OPI introduced in #30.

### DIFF
--- a/examples/exmp046_server/job.py
+++ b/examples/exmp046_server/job.py
@@ -2,6 +2,8 @@
 
 from pathlib import Path
 import shutil
+import sys
+import warnings
 
 from opi.core import Calculator
 from opi.input.structures.structure import Structure
@@ -17,10 +19,11 @@ from opi.external_methods.server import OpiServer, ServerStatus
 # like logging into your HuggingFace account.
 # More information are given in the respective tutorial here:
 # https://www.faccts.de/docs/orca/6.1/tutorials/workflows/extopt.html
-#-------------------------------------------------------------------------
+# -------------------------------------------------------------------------
 
 # Initiate Server object
 server = OpiServer(
+    # Change the following path to a server script, e.g., from orca-external-tools
     serverpath="/path/to/working/server/script/from/orca-external-tools"
 )
 
@@ -28,12 +31,12 @@ server = OpiServer(
 status = server.start_server(exe="bash")
 if status != ServerStatus.RUNNING:
     if status == ServerStatus.ALREADY_RUNNING:
-        print("Some server was already running. Using old one.")
+        warnings.warn("Some server was already running. Using old one.")
     elif status == ServerStatus.PORT_IN_USE:
-        print("Port for server in use. Continue on your own risk.")
+        warnings.warn("Port for server in use. Continue on your own risk.")
     else:
         print(f"Server setup returned with error {status}")
-        exit(1)
+        sys.exit(1)
 
 # Working dir handling
 working_dir = Path("RUN")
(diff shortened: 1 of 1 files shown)


examples/exmp046_server/job.py
```diff
@@ -0,0 +1,59 @@
...
+    serverpath="/path/to/working/server/script/from/orca-external-tools"
+)
+
+# Start server
+status = server.start_server(exe="bash")
+if status != ServerStatus.RUNNING:
+    if status == ServerStatus.ALREADY_RUNNING:
+        print("Some server was already running. Using old one.")
+    elif status == ServerStatus.PORT_IN_USE:
+        print("Port for server in use. Continue on your own risk.")
```
haneug: Here you could use a warning

examples/exmp046_server/job.py
```diff
@@ -0,0 +1,59 @@
...
+    else:
+        print(f"Server setup returned with error {status}")
+        exit(1)
```
haneug: exit(1) should only be used in interactive python. Instead one should use sys.exit(1) but you could also raise an exception like `RuntimeError`

examples/exmp046_server/job.py
```diff
@@ -0,0 +1,59 @@
...
+
+# Working dir handling
+working_dir = Path("RUN")
+shutil.rmtree(working_dir, ignore_errors=True)
+working_dir.mkdir()
+
+# Calculator setup
+structure = Structure.from_xyz("h2o.xyz")
+orca_calc = Calculator(basename="uma_ext", working_dir=working_dir)
```
haneug: here the basename is uma but you did not really specify a script to use. Maybe just call it `extopt`?

examples/exmp046_server/job.py
```diff
@@ -0,0 +1,59 @@
...
+# You can get them from https://github.com/faccts/orca-external-tools
+# Remember that some scripts like umaserver require additional steps
+# like logging into your HuggingFace account.
+# More information are given in the respective tutorial here:
+# https://www.faccts.de/docs/orca/6.1/tutorials/workflows/extopt.html
+#-------------------------------------------------------------------------
+
+# Initiate Server object
+server = OpiServer(
+    serverpath="/path/to/working/server/script/from/orca-external-tools"
```
haneug, line 27: Maybe to make it more obvious write here that this Path has to be changed to a server script of orca-external-tools.

examples/exmp046_server/job.py
```diff
@@ -0,0 +1,59 @@
...
+shutil.rmtree(working_dir, ignore_errors=True)
+working_dir.mkdir()
+
+# Calculator setup
+structure = Structure.from_xyz("h2o.xyz")
+orca_calc = Calculator(basename="uma_ext", working_dir=working_dir)
+orca_calc.structure = structure
+orca_calc.input.ncores = 2
+orca_calc.input.add_simple_keywords(ExternalTools.EXTOPT, Opt.OPT)
+path_client = "/path/to/working/client/script/from/orca-external-tools"
```
haneug, line 53: Maybe to make it more obvious write here that this Path has to be changed to a client script of orca-external-tools.

examples/exmp046_server/job.py
cplett: Good idea

examples/exmp046_server/job.py
cplett: Good catch. Thaks

examples/exmp046_server/job.py
cplett: Thanks for pointing that out.

haneug (approved):
Nice :)
We are working on integrating the examples as tests. This example would be required to be excluded or put in a extopt group. But that is nothing you have to worry about now.

PR #134, opened 2025-09-10
Fixing some Windows compatibility issues

timmyte:
* The OS type check for Windows in `opi/utils/misc.py` was faulty, would never return `True`.
Added the routines `get_os_name()` and `is_os()` that streamline getting and checking the OS type.
* `_determine_orca_paths()`: Check of variant 1 would fail on Windows due to missing `.exe` in binary name.
Added new routine `resolve_binary_name()` that handles the `.exe` on Windows consistently.


src/opi/lib/orca_binary.py
```diff
@@ -0,0 +1,20 @@
...
+    """
+    List of relevant ORCA binaries.
+    """
+
+    ORCA = "orca"
+    ORCA_2JSON = "orca_2json"
+    ORCA_2AIM = "orca_2aim"
+    ORCA_2MKL = "orca_2mkl"
+    ORCA_CRYSTALPREP = "orca_crystalprep"
+    ORCA_JFT = "orca_lft"
```
haneug: Is this a typo? Should it be `ORCA_LFT` ?

timmyte: Eyes like an eagle :+1:

PR #137, opened 2025-09-15
Example Tests

nakul680:
Added example tests as discussed 
A lot of the tests it is simply checked whether ORCA terminated normally , which is already done in the example, so not sure it is needed . 
Also changed the conftest.py file because that was the only way to get the tests working for me , please let me know if they break anything when testing.
Edited all examples to be functions.
Also includes a single core test I wrote a while back , not sure if this is still required.
TODO:
- Exploring the option of having a single script that runs all examples, after which we could have separate examples to check specific values, had some problems with the removal of working directory after every test , but if you want I can look into that 
- Have to look into a way to separate this from normal tests since these tests require ORCA and should not be run on the server.

### DIFF
--- a/src/opi/core.py
+++ b/src/opi/core.py
@@ -176,20 +176,38 @@ class Calculator:
     # &&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&
     # METHODS
     # &&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&
-    def write_input(self) -> None:
+    def write_input(self, force: bool = True) -> bool:
         """
         Function to create the ORCA input file `.inp`.
+
+        Parameters
+        -----------
+        force : bool, default: True
+            Whether to overwrite the ORCA input file if it already exists.
 
         Raises
         ------
         RuntimeError
           * When `.inp` cannot be written.
+          * When '.inp' file already exists and force is `False`.
         ValueError
           * When the `moinp` path is given, and it is not a subpath of the working directory.
+
+        Returns
+        -------
+        bool
+            Whether an existing ORCA .inp file was overwritten.
         """
 
         assert self.working_dir
         self._inpfile = self.working_dir / f"{self.basename}.inp"
+
+        exists = self._inpfile.exists()
+        if exists and not force:
+            raise RuntimeError(
+                f"Input file {self._inpfile} already exists and cannot be overwritten."
+            )
+        input_overwritten = exists and force
 
         # add JSON generation to output blocks
         if self.json_via_input:
@@ -272,6 +290,8 @@ class Calculator:
                         if item.pos is ArbitraryStringPos.BOTTOM:
                             inp.write(f"\n{item}\n")
 
+                return input_overwritten
+
         except IOError as err:
             raise RuntimeError(
                 # Raises an error if the input file cannot be written
@@ -300,7 +320,7 @@ class Calculator:
         """Create a `Runner` object passing on `self.working_dir`."""
         return Runner(working_dir=self.working_dir)
 
-    def run(self, *, timeout: int = -1) -> None:
+    def run(self, *, timeout: int = -1) -> bool:
         """
         Execute ORCA calculation.
 
@@ -309,10 +329,17 @@ class Calculator:
         timeout : int, default: = -1
             Timeout in seconds to wait for ORCA process.
             If value is smaller than zero, wait indefinitely.
+
+        Returns
+        -------
+        bool
+            Whether the ORCA calculation terminated normally.
         """
         runner = self._create_runner()
         assert self.inpfile
         runner.run_orca(self.inpfile, timeout=timeout)
+        output = self.get_output()
+        return output.terminated_normally()
 
     def create_jsons(self, *, force: bool = False) -> None:
         """
@@ -354,3 +381,23 @@ class Calculator:
             runner.check_version(ignore_errors=False)
         except RuntimeError:
             raise
+
+    def write_and_run(self, force: bool = True, timeout: int = -1) -> bool:
+        """
+        Write ORCA .inp file and execute the ORCA calculation.
+
+
+        Parameters
+        ----------
+        force: bool, default:True
+            Whether to overwrite the ORCA input file if it already exists.
+        timeout: int, default: -1
+            Timeout in seconds to wait for ORCA process.
+
+        Returns
+        --------
+        bool
+            Whether the ORCA calculation terminated normally.
+        """
+        self.write_input(force=force)
+        return self.run(timeout=timeout)


haneug (dismissed):
I made some changes to address some of the issues you mentioned. Please merge 60dabe2f7af250c0314fa7c9b7dd34f76c0d2990

EDIT: I merged the commit. I think from this it should be easier to continue.

examples/exmp001_scf/job.py
```diff
@@ -17,24 +18,24 @@
 from opi.input.simple_keywords import Task
 from opi.input.structures import Structure
 
-if __name__ == "__main__":
-    wd = Path("RUN")
+
+def run_exmp001() -> Output:
+    """Perform a HF/def2-SVP single-point"""
+    current_folder = Path(__file__).parent
```
timmyte: Almost! The current folder would be `Path.cwd()`. Could you think of a better name for the variable?

examples/exmp001_scf/job.py
```diff
@@ -58,8 +59,10 @@
...
     # > is (for this calculation) equal to
     print(output.results_properties.geometries[0].single_point_data.finalenergy)
-    # > is (for this calculation) equal to
-    print(
-        output.results_properties.geometries[0].energy[0].totalenergy[0][0]
-        + output.results_properties.geometries[0].vdw_correction.vdw
-    )
+
+    return output
+
```
timmyte, line 61: One line too much.
@nakul680 
Just run `ruff format --line-length 100` over all examples. That should fix the issue.

@haneug 
Can you make sure that we also run `ruff format` over the examples automatically.

examples/exmp002_scf_ccsdt/job.py
```diff
@@ -37,3 +42,10 @@
     print(output.results_properties.geometries[0].single_point_data.finalenergy)
     print("Correlation energy")
     print(output.results_properties.geometries[0].energy[1].correnergy[0][0])
+
+    return output
+
```
timmyte, line 47: One blank line too much.
See above: https://github.com/faccts/opi/pull/137/files#r2358210419

examples/exmp001_scf/job.py
```diff
@@ -17,24 +18,24 @@
 from opi.input.simple_keywords import Task
 from opi.input.structures import Structure
 
-if __name__ == "__main__":
-    wd = Path("RUN")
+
+def run_exmp001() -> Output:
+    """Perform a HF/def2-SVP single-point"""
```
timmyte: I would put the docstring here at the very top of the module file, there it has the best visibility.

@haneug 
IMO we can generally refrain from documenting the main routines in every example, as the structure is always the same. The only exception being example specific helper functions.
It's better to outline the general structure of an example once somewhere in the documentation or in some `README` in this folder.

examples/exmp035_moplot/job.py
```diff
@@ -67,11 +60,14 @@
     print(density)
     print(spin_density)
 
-    # > save mo in script dir
-    with open(mo_5.path.name,"w") as file:
-        file.write(mo_5.cube)
-    # > save mo in script dir line by line
-    with open(f"{mo_5.path.name}.from_iterator", "w") as file:
+    # > save mo in working dir line by line
+    with open(current_folder / f"{mo_5.path.name}.from_iterator", "w") as file:
```
timmyte: Is the MO file name actually supposed to end with `.from_iterator`? 
Don't MO files have some reasonable common file suffix?

tests/examples/test_exmp001_scf.py
```diff
@@ -0,0 +1,17 @@
+import pytest
```
timmyte, line 1: @haneug 
We need some documentation and guidelines for test of the examples.
@nakul680 did a good initial job. The tests look solid.

But it's not explicitly clear:
* How many tests per example?
    * We should agree if a single test can only have a single `assert` or multiple? 
* What should the test exactly test for (this would also depend on the number of tests per example)?
    * Some of these test have 9 `asserts`. A bit more documentation per test would be great.

tests/examples/test_exmp001_scf.py
```diff
@@ -0,0 +1,17 @@
...
+
+from examples.exmp001_scf.job import run_exmp001
+
+
+@pytest.mark.examples
+@pytest.mark.orca
+def test_exmp001_scf(cleanup_run) -> None:
```
timmyte: Type annotation for function parameter missing.

tests/examples/test_exmp004_freq.py
```diff
@@ -0,0 +1,18 @@
+import pytest
+
+from examples.exmp004_freq.job import run_exmp004
+
+
+@pytest.mark.examples
+@pytest.mark.orca
+def test_exmp004_freq(cleanup_run):
```
timmyte: Type annotation for function and return value missing.

tests/examples/test_exmp004_freq.py
```diff
@@ -0,0 +1,18 @@
...
+    output = run_exmp004()
+    assert output.terminated_normally()
```
timmyte: Here especially, more documentation is required.

tests/fixtures/cleanup_run.py
```diff
@@ -0,0 +1,24 @@
+import shutil
+from pathlib import Path
+
+import pytest
+
+
+@pytest.fixture
+def cleanup_run(request):
```
timmyte: Please add a docstring and type annotations.

tests/fixtures/cleanup_run.py
```diff
@@ -0,0 +1,24 @@
...
+    yield
+    run_path = Path(__file__).parent.parent / "RUN"
+    if run_path.exists():
```
timmyte: Please add comments describing what this if-construct's purpose is.

tests/fixtures/cleanup_run.py
```diff
@@ -0,0 +1,24 @@
...
+        if getattr(request.node, "rep_call", None) and request.node.rep_call.passed:
+            shutil.rmtree(run_path)
+            # shutil.rmtree(str(run_path))
+        else:
+            print(f"\nTest failed , RUN directory of failed test {run_path}")
+
+
+@pytest.hookimpl(hookwrapper=True)
+def pytest_runtest_makereport(item, call):
+    """Attach test outcome reports (setup/call/teardown) to the item object."""
```
timmyte: Please describe why this is necessary.

pyproject.toml
```diff
@@ -115,6 +115,14 @@ docs = [
 xfail_strict = true
 strict_markers = true
 
+[tool.pytest.ini_options]
+markers = [
```
timmyte, line 119: Uhhh. That a very nice feature :D

tests/conftest.py
```diff
@@ -10,7 +10,9 @@
...
 # >> Searching for Python modules which do no start with an underscore and converting file path to module path.
 pytest_plugins = [
-    str(filename).removesuffix(".py").replace("/", ".")
-    for filename in Path("tests/fixtures").glob("*.py")
+    #    str(filename).removesuffix(".py").replace("/", ".")
+    f"tests.fixtures.{filename.stem}"
+    for filename in Path(__file__).parent.joinpath("fixtures").glob("*.py")
     if not filename.name.startswith("_")
 ]
+print(pytest_plugins)
```
timmyte: Is that still debug code?

tests/examples/test_exmp001_scf.py
nakul680: That is one of my concerns as well, I'm not sure what exactly I should check for so I just checked for everything with the plan to cut them down after @haneug review , after which I would document them.

examples/exmp001_scf/job.py
haneug: I included the examples folder in the ruff include. This way also linting and import checks/sorting is performed which is reasonable I think.

examples/exmp001_scf/job.py
haneug: I added a `README` where I will next document the tests and what they do.

tests/examples/test_exmp001_scf.py
haneug: I will go through the tests and update the assert statements. For example tests I would allow multiple assert statements while for unit tests only a single assert statement should be used at a time.

tests/examples/test_exmp001_scf.py
haneug: As discussed not for the tests

haneug:
The documentation is now improved but of course still not optimal. The most important part for me is that one can use pytest to run all examples and verify that they run through which is possible with this PR.

examples/README.md
```diff
@@ -0,0 +1,69 @@
...
+
+- **Learning resource:** They show new users how to set up and run typical calculations with OPI.
+- **Integration tests:** The same examples are executed by our test suite (see [`tests/examples`](../tests/examples)) to ensure that OPI's basic functionality and ORCA interface continue to work.
+
+## Notes for Users
+
+- To run an example manually, you can execute the corresponding script within the folder:
+
+```bash
+python example001.py
```
timmyte: I guess you meant
```suggestion
python3 job.py
```

examples/README.md
```diff
@@ -0,0 +1,69 @@
...
+- exmp013_bs: Perform a broken symmetry calculation with B3LYP/SVP (B3LYP/G)
+- exmp014_led: Perform an LED decomposition with DLPNO-CCSD(T)/cc-pVDZ
+- exmp015_pop_analysis: Get multiple population analysis results from DSD-PBEP86/def2-SVP
+- exmp016_autoci: Use AUTO-CI to run CISD/def2-SVP single-point
+- exmp017_roci: Perform a ROCIS calculation
+- exmp018_cipsi: Perform a ICE-CI calculation
+- exmp019_engrad: Perfrom an energy & gradient calculation with r²SCAN-3c
+- exmp020_smd: Run a r²SCAN-3c+SMD(Water) calculation
+- exmp021_basis: Run a BP86/def2-SVP energy calculation with additional diffuse function for oxygen
+- exmp022_scf: Run a PBE0/def2-SVP energy calculation and rotate the initial SCF guess
```
timmyte: Please rename folder to `exmp022_scf_block`.

examples/README.md
```diff
@@ -0,0 +1,69 @@
...
+- exmp023_thermo: Obtain thermostatistical corrections from a frequency calculation
+- exmp024_blocks: Demonstrate input of blocks
+- exmp025_goat: Perform a GOAT conformer search
+- exmp026_scan: Perform scans of bond/angle/dihedral
+- exmp027_constraints: Perform a constrained geometry optimization
+- exmp028_nevpt2: Run a NEVPT2 calculation
+- exmp029_oomp2: Run a OO-MP2 calculation
+- exmp030_eom: Run a EOM-CCSD calculation
+- exmp031_dlpno: Perform a DLPNO-CCSD(T) calculation
```
timmyte: Please rename to `exmp031_dlpno_cc` or `exmp031_dlpno_ccsdt`

tests/conftest.py
```diff
@@ -5,12 +5,71 @@
...
+        # > if `tmp_path` is in the functions signature
+        tmp = cast(Optional[Path], funcargs.get("tmp_path"))
+        if tmp is not None:
+            # > make mypy happy by making sure `when` is not None
+            when = rep.when or "call"
+            item.add_report_section(when, "scratch", str(tmp))
+            print(f"[scratch-dir] {tmp}")
+
+
+@pytest.fixture
```
timmyte: Please put fixtures into `fixtures/`.
Otherwise, very good idea to introduce these routines ;)

tests/conftest.py
```diff
@@ -5,12 +5,71 @@
...
 pytest_plugins = [
-    str(filename).removesuffix(".py").replace("/", ".")
-    for filename in Path("tests/fixtures").glob("*.py")
+    f"tests.fixtures.{filename.stem}"
+    for filename in Path(__file__).parent.joinpath("fixtures").glob("*.py")
     if not filename.name.startswith("_")
 ]
+
+
+# > hookwrapper for printing the scratch directory when a test with `tmp_path` in its signature fails
```
timmyte: Please add an example to the docstring how this would, so one knows what to look for.

examples/exmp001_scf/job.py
```diff
@@ -5,39 +5,37 @@
 from pathlib import Path
```
timmyte, line 5: Please put an empty `__init__.py` into the `tests/examples/` folder.

tests/examples/test_exmp002_scf_ccsdt.py
```diff
@@ -0,0 +1,22 @@
...
+
+    # Get input file from example folder
+    input_file = example_input_file(run_exmp002)
+    structure = Structure.from_xyz(input_file)
+
+    # Run the example in tmp_path
+    output = run_exmp002(structure=structure, working_dir=tmp_path)
+
+    # Assert final energy
+    assert output.get_final_energy()
```
timmyte: Most assert statements have no condition. Please add these for quantities where it matters if its positive or negative

PR #149, opened 2025-09-30
Introduce multiple Python versions to nox run

haneug:
We want to use multiple Python versions in nox run. I added two new constants `PYTHON_VERSIONS` and `SINGLE_PYTHON_VERSION` that hold the Python versions that should be used. Right now multiple Python versions are only used for mypy (linting, code formatting, sorting and removing of imports should be version independent I think, please correct me if I am wrong). The sessions which should be run with multiple Python versions are type checking, unit tests and integration/example tests. For the example tests I did not add multiple Python version yet, because ORCA calculations are expensive.

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -4,6 +4,11 @@ import os
 # > External packages
 import nox
 
+# > Sessions that run with a single Python version run the following version
+SINGLE_PYTHON_VERSION = ["3.11"]
+# > Sessions that run with multiple Python versions run the following versions
+PYTHON_VERSIONS = ["3.11", "3.12", "3.13"]
+
 # > Don't update lock file when setting up virtual envs.
 os.environ.update({"UV_FROZEN": "1"})
 # > Disable automatic donwload of Python distributions
@@ -39,7 +44,7 @@ nox.options.default_venv_backend = "uv"
 # //////////////////////////////////////////
 # ///         UNIT TESTS: pytest         ///
 # //////////////////////////////////////////
-@nox.session(default=False)
+@nox.session(default=False, python=SINGLE_PYTHON_VERSION)
 def tests(session):
     session.run_install(
         "uv",
@@ -54,7 +59,7 @@ def tests(session):
 # //////////////////////////////////////////
 # ///     STATIC TYPE CHECKING: mypy     ///
 # //////////////////////////////////////////
-@nox.session(tags=["static_check", "pr_check"])
+@nox.session(tags=["static_check", "pr_check"],python=PYTHON_VERSIONS)
 def type_check(session):
     session.run_install(
         "uv",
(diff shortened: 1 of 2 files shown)


noxfile.py
```diff
@@ -4,6 +4,13 @@
 # > External packages
 import nox
 
+# > Sessions that run with multiple Python versions run the following versions
+PYTHON_VERSIONS = ["3.11", "3.12", "3.13"]
+
+def running_in_ci() -> bool:
```
timmyte: This is not ideal.
The noxfile should be agnostic to the environment it's executed in and not contain functions that are specific to one.
Please look for another way to tell Nox what Pyton versions to test or not to test. The information should come from the workflow file directly.

examples/exmp042_element/job.py
```diff
@@ -20,7 +20,7 @@ def run_exmp042(
     if structure is None:
         structure = Structure.from_xyz("inp.xyz")
 
-    calc = Calculator(basename="job", working_dir=working_dir)
+    calc = Calculator(basename="job", working_dir=working_dir, version_check=False)
```
timmyte, line 24: Why disable the version check?

noxfile.py
```diff
@@ -39,7 +46,7 @@
 # //////////////////////////////////////////
 # ///         UNIT TESTS: pytest         ///
 # //////////////////////////////////////////
-@nox.session(default=False)
+@nox.session(default=False,python=None if running_in_ci() else PYTHON_VERSIONS)
```
timmyte: 1)
Don't reuse expressions. Store it in a module variable and use that instead.

2)
```suggestion
@nox.session(default=False, python=None if running_in_ci() else PYTHON_VERSIONS)
```

noxfile.py
```diff
@@ -54,7 +61,7 @@ def tests(session):
 # //////////////////////////////////////////
 # ///     STATIC TYPE CHECKING: mypy     ///
 # //////////////////////////////////////////
-@nox.session(tags=["static_check", "pr_check"])
+@nox.session(tags=["static_check", "pr_check"],python=None if running_in_ci() else PYTHON_VERSIONS)
```
timmyte:
```suggestion
@nox.session(tags=["static_check", "pr_check"], python=None if running_in_ci() else PYTHON_VERSIONS)
```

timmyte (changes requested):
Nice idea to test multiple version in one session ;)

examples/exmp042_element/job.py
haneug: This is done to be ORCA independent, otherwise the version check will be triggered and fail if ORCA is not available. I added a comment about this.

noxfile.py
haneug: I will remove the python versions from the noxfile. The yaml file controls the python version used in the CI.

noxfile.py
haneug: resolved with Python version removal

noxfile.py
haneug: resolved with Python version removal

haneug:
I now removed the Python versions from the noxfile since we want to drive the python version used from the outside, e.g., see workflow file.

PR #151, opened 2025-10-06
WIP: Improve Docker usage

haneug:
Since the docker does not write a gbw file and the property file is also almost empty we want to obtain the results from multi-xyz files. Here I added an example on how to extract the Structures.

ToDO:
- [ ] Add the option to parse energies from the xyz files. These should not be part of the Structures as such data can change but of a seperate properties class.


src/opi/input/blocks/block_docker.py
```diff
@@ -13,22 +13,41 @@ class BlockDocker(Block):
 
     _name: str = "docker"
     docklevel: Literal["screening", "normal", "quick", "complete"] | None = None
+    """Literal: "screening", "normal", "quick", "complete". Defines docking strategy."""
```
timmyte, line 16: Thanks for documenting ;)

src/opi/input/structures/properties.py
```diff
@@ -0,0 +1,321 @@
...
+from io import StringIO
+from os import PathLike
+from pathlib import Path
+from typing import Literal, Sequence
+
+from opi.utils.textio import TrackingTextIO
+
+__all__ = ("Properties",)
+
+RGX_INT_AND_FLOAT = re.compile(r"[+-]?(?:\d+\.\d*|\.\d+|\d+)(?:[Ee][+-]?\d+)?")
```
timmyte: Good start for an integer and float regex, but this one, will extract every single digit even from within a word, e.g. `high5low42ef -> 5, 42`.
Please make sure to add boundary conditions.

src/opi/input/structures/properties.py
```diff
@@ -0,0 +1,321 @@
...
+
+class Properties:
+    """
+    Class for keeping additional Structure properties found, e.g., in an xyz file.
+
+    Attributes
+    ----------
+    structure_id : int
+        Number of the structure from which the properties are.
+    energy_total : float
```
timmyte: Can also be and is by default `None`. Please update docstring.

src/opi/input/structures/properties.py
```diff
@@ -0,0 +1,321 @@
...
+        Energy of a structure.
+    energy_relative : float
```
timmyte: Can also be and is by default `None`. Please update docstring.

src/opi/input/structures/properties.py
```diff
@@ -0,0 +1,321 @@
...
+        self.energy_relative: float | None = energy_relative
+
+    @classmethod
+    def from_xyz(
+        cls, xyzfile: Path | str | PathLike[str], mode: Literal["goat", "docker"] = "goat"
+    ) -> "Properties":
+        """
+        Function for reading xyz file and return a Properties object.
+
+        Parameters
```
timmyte, line 50: Please document `mode`.

src/opi/input/structures/properties.py
```diff
@@ -0,0 +1,321 @@
...
+        Raises
+        --------
+        FileNotFoundError
+            If the XYZ file cannot be found
+        ValueError
+            If there is a problem with parsing the XYZ file
+
+        Returns
+        --------
+        `Properties`:`Properties object extracted from file
```
timmyte: Don`t need the backticks here.

examples/exmp049_docker/job.py
```diff
@@ -0,0 +1,68 @@
...
+    if not output.terminated_normally():
+        print(f"ORCA calculation failed, see output file: {output.get_outfile()}")
+        sys.exit(1)
+    # << END OF IF
+
+    # > Parse JSON files
+    output.parse()
+
+    # > Print graph to visualize available results
+    print(output.print_graph())
```
timmyte: This is a bit misleading.
Why does the output of `print_xxx` need to be passed into `print()` to be actually printed? Either adjust the name of `print_graph()` or the method itself.

src/opi/input/structures/properties.py
```diff
@@ -0,0 +1,321 @@
...
+        Parameters
+        ----------
+        trj_file : Path | str | PathLike[str]
+            Name or path to xyz file with multiple structures
+        comment_symbols: str | Sequence[str] | None, default: None
+            List of symbols that indicate user comments in the xyz file. User comments are skipped before the actual xyz
+            data starts. By default, no user comments are used. White-space only comments are not allowed and are
+            silently ignored.
+        struc_limit: int | None, default: None
+            Limit of structures that should be considered from the trj.xyz file. With the default, None, all structures are
```
timmyte: >  Limit of structures that should be considered from the trj.xyz file.

The sentence as well as the parameter name is weird. How about: 
"nstructures: If >0, only read the first n structures."

* I would also suggest, `0` and `None` have the same meaning.
* And negative numbers just count from the end, i.e. return all entries except the last x.

src/opi/input/structures/properties.py
```diff
@@ -0,0 +1,321 @@
...
+            Limit of structures that should be read from the trj.xyz string. With the default, None, all structures are
+            read.
+
+        Returns
+        --------
+        `list[Properties]`:`Properties objects extracted from file
+        """
+        properties_list: list[Properties] = []
+
+        with StringIO(trj_string) as f_xyz:
```
timmyte:
```suggestion
        with TrackingTextIO(trj_string) as f_xyz:
```
would that also be possible? If yes, please change.

src/opi/input/structures/properties.py
```diff
@@ -0,0 +1,321 @@
...
+                        tracked, comment_symbols=comment_symbols, mode=mode
+                    )
+                    if properties is None:
+                        break
+                    properties_list.append(properties)
+                except ValueError:
+                    raise
+                n_struc += 1
+                # > Break if limit of allowed structures is reached
+                if struc_limit is not None and n_struc >= struc_limit:
```
timmyte: This is redundant code. Even if it's only a few lines, you copied the whole logic of `struc_limit`. Please construct the functions such, that redundancy is minimal.

src/opi/input/structures/properties.py
```diff
@@ -0,0 +1,321 @@
...
+
+        Parameters
+        ----------
+        trj_string : Path | str | PathLike[str]
+            String that contains multiple xyz file data (trajectory data)
+        comment_symbols: str | Sequence[str] | None, default: None
+            List of symbols that indicate user comments in the xyz file. User comments are skipped before the actual xyz
+            data starts. By default, no user comments are used. White-space only comments are not allowed and are
+            silently ignored.
+        struc_limit: int | None, default: None
```
timmyte: See my comment above, about `struc_limit`.

src/opi/input/structures/properties.py
```diff
@@ -0,0 +1,321 @@
...
+    def from_xyz_buffer(
+        cls,
+        xyz_lines: TrackingTextIO,
+        /,
+        *,
+        comment_symbols: str | Sequence[str] | None = None,
+        mode: Literal["goat", "docker"] = "goat",
+    ) -> "Properties | None":
+        """
+        Function for reading from the comment line of a xyz file from a buffer and converting it to a Properties object.
```
timmyte: Reading _what_ from the comment line?

src/opi/input/structures/properties.py
```diff
@@ -0,0 +1,321 @@
...
+        cls,
+        trj_string: str,
+        /,
+        *,
+        mode: Literal["goat", "docker"] = "goat",
+        comment_symbols: str | Sequence[str] | None = None,
+        struc_limit: int | None = None,
+    ) -> "list[Properties]":
+        """
+        Function for reading multi-xyz data from string and returning a Properties object.
```
timmyte: 1) Terminology is inconsistent. Stick with 'multi-xyz' or 'trajectory-xyz'.
2) What exactly is read from the "multi-xyy data". Be more precise.

src/opi/input/structures/properties.py
```diff
@@ -0,0 +1,321 @@
...
+
+
+class Properties:
+    """
+    Class for keeping additional Structure properties found, e.g., in an xyz file.
+
+    Attributes
+    ----------
+    structure_id : int
+        Number of the structure from which the properties are.
```
timmyte: Does this correspond to an actual attribute of the `Structure` class, if yes, mention it.
Do these "numbers" start from 0 or 1?

src/opi/input/structures/properties.py
```diff
@@ -0,0 +1,321 @@
...
+from opi.utils.textio import TrackingTextIO
+
+__all__ = ("Properties",)
+
+RGX_INT_AND_FLOAT = re.compile(r"[+-]?(?:\d+\.\d*|\.\d+|\d+)(?:[Ee][+-]?\d+)?")
+
+
+class Properties:
+    """
+    Class for keeping additional Structure properties found, e.g., in an xyz file.
```
timmyte: What would "non-additional" properties be? 
I would just drop that word.

src/opi/input/structures/properties.py
```diff
@@ -0,0 +1,321 @@
...
+
+    Attributes
+    ----------
+    structure_id : int
+        Number of the structure from which the properties are.
+    energy_total : float
+        Energy of a structure.
+    energy_relative : float
+        Relative energy of a structure.
```
timmyte: Relative to what? Mention that the reference energy can be any.

src/opi/input/structures/properties.py
```diff
@@ -0,0 +1,321 @@
...
+        self.structure_id: int | None = structure_id
+        self.energy_total: float | None = energy_total
+        self.energy_relative: float | None = energy_relative
+
+    @classmethod
+    def from_xyz(
+        cls, xyzfile: Path | str | PathLike[str], mode: Literal["goat", "docker"] = "goat"
+    ) -> "Properties":
+        """
+        Function for reading xyz file and return a Properties object.
```
timmyte: Where are these properties stored?
Is the following also supported by this reader:
```
3
         [GRADIENT]
H 0 0 0   8 9 7
H 0 0 0   2 5 6
```

src/opi/input/structures/properties.py
```diff
@@ -0,0 +1,321 @@
...
+        comment_symbols: str | Sequence[str] | None = None,
+        struc_limit: int | None = None,
+    ) -> "list[Properties]":
+        """
+        Function for reading multi-xyz file and returning a Properties object.
+
+        Parameters
+        ----------
+        trj_file : Path | str | PathLike[str]
+            Name or path to xyz file with multiple structures
```
timmyte: What if the trajectory would only contain a single XYZ entry?

src/opi/input/structures/properties.py
```diff
@@ -0,0 +1,321 @@
...
+        Raises
+        --------
+        FileNotFoundError
+            If the XYZ file cannot be found
+        ValueError
+            If there is a problem with parsing the XYZ file
+
+        Returns
+        --------
+        `list[Properties]`:`Properties object extracted from file
```
timmyte: Don`t need backticks.

src/opi/input/structures/properties.py
```diff
@@ -0,0 +1,321 @@
...
+            List of symbols that indicate user comments in the xyz file. User comments are skipped before the actual xyz
+            data starts. By default, no user comments are used. White-space only comments are not allowed and are
+            silently ignored.
+        struc_limit: int | None, default: None
+            Limit of structures that should be read from the trj.xyz string. With the default, None, all structures are
+            read.
+
+        Returns
+        --------
+        `list[Properties]`:`Properties objects extracted from file
```
timmyte: Backticks obsolete.

src/opi/input/structures/properties.py
```diff
@@ -0,0 +1,321 @@
...
+        Function for reading from the comment line of a xyz file from a buffer and converting it to a Properties object.
+
+        Parameters
+        ----------
+        xyz_lines: TrackingTextIO
+            A buffer that contains xyz file data
+        comment_symbols: str | Sequence[str] | None, default: None
+            List of symbols that indicate user comments in the xyz file. User comments are skipped before the actual xyz
+            data starts. By default, no user comments are used. White-space only comments are not allowed and are
+            silently ignored.
```
timmyte: `mode` is not mentioned here.

src/opi/input/structures/properties.py
```diff
@@ -0,0 +1,321 @@
...
+        comments_tuple: tuple[str, ...] | None = None
+
+        # > Convert comments to tuple
+        if isinstance(comment_symbols, str):
+            comments_tuple = (comment_symbols,)
+        elif isinstance(comment_symbols, Sequence):
+            comments_tuple = tuple(comment_symbols)
+
+        # > Skip arbitrary number of empty and user comment lines at the beginning
+        while (line := xyz_lines.readline()) != "":
```
timmyte: Please pick one style to look for an empty line, either `[not] line` or `line !=/== ""`. Don't intermix them.

src/opi/input/structures/properties.py
```diff
@@ -0,0 +1,321 @@
...
+            line = xyz_lines.readline()
+            # > empty lines are not allowed
+            if line == "":
+                raise ValueError(f"Line {xyz_lines.line_number}: Incomplete xyz file buffer")
+
+        return properties
+
+    @classmethod
+    def docker_energies(cls, line: str) -> "Properties":
+        """Function for reading docker energies from string and return a Properties object."""
```
timmyte: 1)
```suggestion
        """Function for reading DOCKER energies from string and return a Properties object."""
```

2)
From any string? Please document what the string has to look like.

src/opi/input/structures/properties.py
```diff
@@ -0,0 +1,321 @@
...
+        # > Parse the comment line
+        try:
+            # > ID of the structure
+            structure_id = int(numbers[0])
+            # > Total energy is second number (Eh)
+            energy_total = numbers[1]
+            # > Relative energy is the third number (kcal/mol)
+            energy_relative = numbers[2]
+        except IndexError:
+            raise ValueError("Could not parse docker energies from comment line.")
```
timmyte:
```suggestion
        except (IndexError, ValueError) as err:
            raise ValueError("Could not parse docker energies from comment line.") from err
```

src/opi/input/structures/properties.py
```diff
@@ -0,0 +1,321 @@
...
+        properties = Properties(
+            structure_id=structure_id, energy_total=energy_total, energy_relative=energy_relative
+        )
+        return properties
+
+    @classmethod
+    def goat_energies(cls, line: str) -> "Properties":
+        """Function for reading goat energies from string and return a Properties object."""
```
timmyte: 1)

```suggestion
        """Function for reading GOAT energies from string and return a Properties object."""
```
2) From any string? Please document what the string has to look like.

src/opi/input/structures/properties.py
```diff
@@ -0,0 +1,321 @@
...
+        numbers = [float(m) for m in re.findall(RGX_INT_AND_FLOAT, line)]
+        # > Parse the comment line
+        try:
+            energy_total = numbers[0]
+        except IndexError:
+            raise ValueError("Could not parse docker energies from comment line.")
```
timmyte:
```suggestion
        except IndexError as err:
            raise ValueError("Could not parse docker energies from comment line.") from err
```

src/opi/output/core.py
```diff
@@ -321,10 +321,9 @@ def collect_json_files(
         """
         files = []
         if isinstance(pattern, str):
-            # > String pattern must not end with ".json"!
-            gbw_file = self.get_file(pattern + ".json")
+            gbw_file = self.get_file(pattern)
             if gbw_file.is_file():
-                files.append(gbw_file)
+                files.append(gbw_file.with_suffix(".json"))
```
timmyte: Please explain. I don't get what's the purpose of this change.
To my understanding: First we have path to GBW file stored in `gbw_file`, then we check if the GBW file exists. If the file exists, we just **assume** the corresponding JSON file also exist!?

Please correct me if I'm just too dense to understand.

tests/examples/test_exmp025_goat.py
```diff
@@ -14,4 +14,9 @@ def test_exmp025_goat(example_input_file, tmp_path) -> None:
     structure = Structure.from_xyz(input_file)
 
     # Run the example in tmp_path
-    run_exmp025(structure=structure, working_dir=tmp_path)
+    structures, properties_list = run_exmp025(structure=structure, working_dir=tmp_path)
+
+    # Assert that docker structures are present
+    assert len(structures) >= 1
```
timmyte:
```suggestion
    assert structures
```
would be sufficient.

tests/examples/test_exmp050_docker.py
```diff
@@ -0,0 +1,22 @@
...
+    """Run a docker calculation and gather the resulting structures."""
+    # Get input file from example folder
+    input_file = example_input_file(run_exmp049)
+    structure = Structure.from_xyz(input_file)
+
+    # Run the example in tmp_path
+    structures, properties_list = run_exmp049(structure=structure, working_dir=tmp_path)
+
+    # Assert that docker structures are present
+    assert len(structures) >= 10
```
timmyte, line 20: Please mention the "10" in the comment. In the GOAT test we only look if we have any structures at all.

src/opi/input/structures/properties.py
haneug: It is possible after implementing \_\_enter\_\_ and \_\_exit\_\_ to `TrackingTextIO` and allowing string initialization. I will change this accordingly.

src/opi/input/structures/properties.py
haneug: - I would prefer `nstruc_limit` or `n_struc_limit`
- Having 0 and None have the same meaning is a good idea
- negative numbers would require to first read the whole file and slice then and since I do not see a lot of benefit here I will not allow negative numbers at all.

src/opi/input/structures/properties.py
haneug: This is a valid point. I moved the loop into a function that returns an Iterator and used the Iterator instead. I applied the same logic to `Structure

src/opi/input/structures/properties.py
haneug: No this is just something written in the GOAT xyz files. It seems that they start from 1 but I am not sure. This information is not really relevant in my opinion.

src/opi/input/structures/properties.py
haneug: No this is not supported but does ORCA write such a file?

src/opi/input/structures/properties.py
haneug: the you get a list with only one structure. I will update the docstring accordingly.

src/opi/output/core.py
haneug: I have to admit that it is easy to get lost here which is not good! It kind of grew this way:
- In `__init__` (here) we first check if the gbw_files exists and save the names of the corresponding json files
- In parse, by default, we create the missing json files. This can be forced so that they are re-created or suppressed so that no new gbw json is created.

src/opi/input/structures/properties.py
haneug: decided to go with [not] line

examples/exmp049_docker/job.py
haneug: print was redundant

src/opi/input/structures/properties.py
haneug: Added boundaries

src/opi/input/structures/properties.py
timmyte: Then, please explicitly state that this does not refer to any `Structure` in OPI.
That’s what I intuitively thought, at least.

src/opi/input/structures/properties.py
timmyte: At least not that I know.
But there are (unofficial) variants of XYZ that have atomic properties after the coordinates.

src/opi/output/core.py
timmyte: Got it! Thanks!
Please add a note to the beginning of the docstring, that JSON files returned, are only assumed paths and might not exist!

src/opi/input/structures/properties.py
```diff
@@ -0,0 +1,319 @@
...
+from collections.abc import Iterator
+from os import PathLike
+from pathlib import Path
+from typing import Literal, Sequence
+
+from opi.utils.tracking_text_io import TrackingTextIO
+
+__all__ = ("Properties",)
+
+# > RE for finding floats and integers guarded by lookarounds that no letters are next to them
```
timmyte, line 11: I guess, we can assume that the number we look for are always in column of their own, i.e. they are only preceded or succeeded by whitespaces.
Therefore, I would look for whitespaces `\s` and beginning of line `\A` or end of line `\Z`.
If you need help just ping me ;)

src/opi/input/structures/properties.py
```diff
@@ -0,0 +1,319 @@
...
+        self.structure_id: int | None = structure_id
+        self.energy_total: float | None = energy_total
+        self.energy_relative: float | None = energy_relative
+
+    @classmethod
+    def from_xyz(
+        cls, xyzfile: Path | str | PathLike[str], mode: Literal["goat", "docker"] = "goat"
+    ) -> "Properties":
+        """
+        Function for reading properties from the comment lines of a xyz file and return a Properties object.
```
timmyte:
```suggestion
        Function for reading properties from the comment line of single structure from a xyz file from a GOAT or DOCKER run and return a Properties object.
        For multi-structure xyz files refer to `from_trj_xyz()`.
```

src/opi/input/structures/properties.py
```diff
@@ -0,0 +1,319 @@
...
+        Raises
+        --------
+        FileNotFoundError
+            If the XYZ file cannot be found
+        ValueError
+            If there is a problem with parsing the XYZ file
+
+        Returns
+        --------
+        Properties:Properties object extracted from file
```
timmyte:
```suggestion
        Properties: Properties object extracted from file
```

src/opi/input/structures/properties.py
```diff
@@ -0,0 +1,319 @@
...
+        cls,
+        trj_file: Path | str | PathLike[str],
+        /,
+        *,
+        mode: Literal["goat", "docker"] = "goat",
+        comment_symbols: str | Sequence[str] | None = None,
+        n_struc_limit: int | None = None,
+    ) -> "list[Properties]":
+        """
+        Function for reading multi-xyz file and returning a Properties object.
```
timmyte: Please make sure to mention that these are currently DOCKER and GOAT specific functions.

src/opi/input/structures/properties.py
```diff
@@ -0,0 +1,319 @@
...
+        mode: Literal["goat", "docker"], default = "goat"
+            Define how the comment line should be processed, e.g, it is the comment line from a DOCKER or GOAT run.
+        comment_symbols: str | Sequence[str] | None, default: None
+            List of symbols that indicate user comments in the xyz file. User comments are skipped before the actual xyz
+            data starts. By default, no user comments are used. White-space only comments are not allowed and are
+            silently ignored.
+        n_struc_limit: int | None, default: None
+            If >0, only read the first n structures.
+
+        Returns
```
timmyte, line 181: Can it also raise anything?

src/opi/input/structures/properties.py
```diff
@@ -0,0 +1,319 @@
...
+        # > Open file and iterate over structures
+        with TrackingTextIO(trj_file.open()) as tracked:
+            return list(cls._iter_xyz_structures(tracked, comment_symbols, mode, n_struc_limit))
+
+    @classmethod
+    def from_xyz_block(
+        cls, xyz_string: str, mode: Literal["goat", "docker"] = "goat"
+    ) -> "Properties":
+        """
+        Function for reading xyz data from string and return a Properties object.
```
timmyte: I would emphasize that it only supports a single structure.

src/opi/input/structures/properties.py
```diff
@@ -0,0 +1,319 @@
...
+        trj_string: str,
+        /,
+        *,
+        mode: Literal["goat", "docker"] = "goat",
+        comment_symbols: str | Sequence[str] | None = None,
+        n_struc_limit: int | None = None,
+    ) -> "list[Properties]":
+        """
+        Function for reading trajectory data from string and returning a Properties object. What is read depends on the
+        mode Literal.
```
timmyte:
```suggestion
        Function for reading trajectory data from string and returning a Properties object.
```

src/opi/input/structures/properties.py
```diff
@@ -0,0 +1,319 @@
...
+
+# > RE for finding floats and integers guarded by lookarounds that no letters are next to them
+RGX_INT_AND_FLOAT = re.compile(
+    r"(?<![A-Za-z])[+-]?(?:\d+\.\d*|\.\d+|\d+)(?:[Ee][+-]?\d+)?(?![A-Za-z])"
+)
+
+
+class Properties:
+    """
+    Class for keeping structure properties found in xyz files.
```
timmyte:
```suggestion
    Class to represent structure properties.
    Currently properties can only be read from XYZ files created with GOAT or DOCKER.
```

src/opi/input/structures/properties.py
```diff
@@ -0,0 +1,319 @@
...
+        comment_symbols: str | Sequence[str] | None = None,
+        n_struc_limit: int | None = None,
+    ) -> "list[Properties]":
+        """
+        Function for reading multi-xyz file and returning a Properties object.
+
+        Parameters
+        ----------
+        trj_file : Path | str | PathLike[str]
+            Name or path to xyz file with one or multiple structure(s)
```
timmyte:
```suggestion
            Name or path to xyz file with one or multiple structure(s).
```

src/opi/input/structures/properties.py
```diff
@@ -0,0 +1,319 @@
...
+    def from_xyz(
+        cls, xyzfile: Path | str | PathLike[str], mode: Literal["goat", "docker"] = "goat"
+    ) -> "Properties":
+        """
+        Function for reading properties from the comment lines of a xyz file and return a Properties object.
+
+        Parameters
+        ----------
+        xyzfile : Path | str | PathLike[str]
+            Name or path to xyz file
```
timmyte: Please be consistent. Either spell `XYZ` all-caps or all-lower. Currently it's completely random in every docstring.

src/opi/input/structures/properties.py
```diff
@@ -0,0 +1,319 @@
...
+        cls,
+        xyz_lines: TrackingTextIO,
+        /,
+        *,
+        comment_symbols: str | Sequence[str] | None = None,
+        mode: Literal["goat", "docker"] = "goat",
+    ) -> "Properties | None":
+        """
+        Function for reading from the comment line of a xyz file from a buffer and converting it to a Properties object.
+        What is read depends on the `mode` Literal, e.g, total energies from a GOAT xyz file.
```
timmyte:
```suggestion
```

src/opi/input/structures/properties.py
```diff
@@ -0,0 +1,319 @@
...
+        Parameters
+        ----------
+        trj_file : Path | str | PathLike[str]
+            Name or path to xyz file with one or multiple structure(s)
+        mode: Literal["goat", "docker"], default = "goat"
+            Define how the comment line should be processed, e.g, it is the comment line from a DOCKER or GOAT run.
+        comment_symbols: str | Sequence[str] | None, default: None
+            List of symbols that indicate user comments in the xyz file. User comments are skipped before the actual xyz
+            data starts. By default, no user comments are used. White-space only comments are not allowed and are
+            silently ignored.
```
timmyte:
```suggestion
            List of symbols that indicate user comments in the xyz file.
            User comments are skipped before the actual xyz data starts.
            White-space only comments are not allowed and are silently ignored.
```
* Please add line breaks after each sentence.
* Please describe if only whole comment lines are supported or also inline comments.
* What does the second sentence describe? So comments are only allowed before the XYZ data?
* I don't get the last sentence? If it's a comment it should be ignored regardless what it contains!? Or am I wrong.

src/opi/input/structures/structure.py
```diff
@@ -402,11 +402,9 @@ def from_xyz(
         --------
         `Structure`:`Structure object extracted from file
         """
-        structures = cls.from_trj_xyz(
-            xyzfile, charge=charge, multiplicity=multiplicity, struc_limit=1
-        )
-        structure = structures[0]
-        return structure
+        return cls.from_trj_xyz(xyzfile, charge=charge, multiplicity=multiplicity, n_struc_limit=1)[
```
timmyte: This is just plain ugly.

If Ruff formats it like that please store `cls.from_trj_xyz(xyzfile, charge=charge, multiplicity=multiplicity, n_struc_limit=1)` in some variable and return `return variable[0]`.

Or see if you can get into the form used further down:
```
return cls.from_trj_xyz_block(
            xyz_string, charge=charge, multiplicity=multiplicity, n_struc_limit=1
        )[0]
```

src/opi/input/structures/structure.py
```diff
@@ -449,37 +446,23 @@ def from_trj_xyz(
         --------
         `list[Structure]`:`Molecular structure objects extracted from the xyz file
```
```suggestion
        list[Structure]: Molecular structure objects extracted from the xyz file
```

src/opi/output/core.py
```diff
@@ -69,6 +69,13 @@ class Output:
         Basename of the job.
     working_dir: Path
         Optional path to the working directory.
+    gbw_files: list[Path]
+        List of paths to existing gbw files in the working directory.
+    gbw_json_files: list[Path]
+        List of paths to gbw json files that can be created from available gbw files. Files may or may not exist in the
+        working directory.
```
```suggestion
        List of paths to gbw json files that can be created from available gbw files.
        Files may or may not exist in the working directory.
```

src/opi/output/core.py
```diff
@@ -300,7 +308,7 @@ def collect_json_files(
         step: int = 1,
     ) -> list[Path]:
         """
-        Searches for available gbw files according to a pattern in `working_dir` and returns a list of corresponding `.json` file paths.
+        Searches for available gbw files according to a pattern in `working_dir` and returns a list of them
```
```suggestion
        Searches for available gbw files according to a pattern in `working_dir` and returns these as list.
```

src/opi/utils/tracking_text_io.py
```diff
@@ -1,20 +1,53 @@
...
+from typing import Any, Optional, TextIO, Type
 
 
 class TrackingTextIO:
     """Class that keeps track of the line number in TextIO"""
 
-    def __init__(self, stream: TextIO) -> None:
-        self._stream: TextIO = stream
+    def __init__(self, stream: str | TextIO) -> None:
+        self._stream: TextIO
```
timmyte, line 11: It this line required? If yes, please add a comment describing why.

src/opi/utils/tracking_text_io.py
```diff
@@ -1,20 +1,53 @@
-from typing import Any, Iterator, TextIO
+import types
+from io import StringIO
+from typing import Any, Optional, TextIO, Type
 
 
 class TrackingTextIO:
```
timmyte, line 6: Why didn't you derive directly from `TextIO`?

timmyte (changes requested):

src/opi/input/structures/properties.py
haneug: Thanks, that did indeed not make any sense :D

src/opi/utils/tracking_text_io.py
haneug: It is required for mypy. I will add a comment.

src/opi/utils/tracking_text_io.py
haneug: `TextIO` is a class just for static typing and not a real runtime class. Therefore I cannot derive from it.

src/opi/input/structures/properties.py
haneug: This is actually not the case. The docker writes a line like this:
```
11 Eopt=-10.1490069025 (Eh) Einter=-4.968881 (kcal/mol)
```

src/opi/input/structures/properties.py
haneug: We decided to ignore this for now.

src/opi/input/structures/properties.py
```diff
@@ -0,0 +1,322 @@
...
+        cls,
+        trj_file: Path | str | PathLike[str],
+        /,
+        *,
+        mode: Literal["goat", "docker"] = "goat",
+        comment_symbols: str | Sequence[str] | None = None,
+        n_struc_limit: int | None = None,
+    ) -> "list[Properties]":
+        """
+        Function for reading multi-XYZ file and returning a Properties object.
```
timmyte: Please update the first sentence. It not a `Properties` object that is returned!

src/opi/input/structures/properties.py
```diff
@@ -0,0 +1,322 @@
...
+        cls,
+        trj_string: str,
+        /,
+        *,
+        mode: Literal["goat", "docker"] = "goat",
+        comment_symbols: str | Sequence[str] | None = None,
+        n_struc_limit: int | None = None,
+    ) -> "list[Properties]":
+        """
+        Function for reading trajectory data from string and returning a `Properties` object.
```
timmyte: Please update the first sentence. It not a `Properties` object that is returned!

src/opi/input/structures/properties.py
```diff
@@ -0,0 +1,322 @@
...
+
+        Raises
+        --------
+        ValueError
+            When no valid properties can be read from the input buffer or the corresponding structure is incomplete.
+
+        Returns
+        --------
+        Properties | None
+            The `Properties` object extracted from the buffer or None if the buffer was empty.
```
timmyte: This seems to be inconsistent.
If `from_xyz_buffer` returns `None` if the buffer is empty, `from_xyz()` and `from_xyz_block()` should do this as well.

This applies to all `from_...()` methods here. Pick only one approach if the source is empty/non-existent, **either always raise an exception or return a null-object** (`None` and empty list for the `..._trj_...()` methods.).

src/opi/input/structures/properties.py
```diff
@@ -0,0 +1,322 @@
...
+            line = xyz_lines.readline()
+            # > empty lines are not allowed
+            if not line:
+                raise ValueError(f"Line {xyz_lines.line_number}: Incomplete XYZ file buffer")
+
+        return properties
+
+    @classmethod
+    def docker_energies(cls, line: str) -> "Properties":
+        """Function for reading DOCKER energies from comment line of a DOCKER XYZ file."""
```
timmyte:
```suggestion
        """Function for reading DOCKER energies from comment line of a DOCKER XYZ file and return them in `Properties` object."""
```

src/opi/input/structures/properties.py
```diff
@@ -0,0 +1,322 @@
...
+        except (IndexError, ValueError) as err:
+            raise ValueError("Could not parse docker energies from comment line.") from err
+        properties = Properties(
+            structure_id=structure_id, energy_total=energy_total, energy_relative=energy_relative
+        )
+        return properties
+
+    @classmethod
+    def goat_energies(cls, line: str) -> "Properties":
+        """Function for reading GOAT energies from comment line of a GOAT XYZ file."""
```
timmyte:
```suggestion
        """Function for reading GOAT energies from comment line of a GOAT XYZ file and return them in `Properties` object."""
```

src/opi/input/structures/structure.py
```diff
@@ -526,15 +509,15 @@ def from_trj_xyz_block(
         charge: int = 0,
         multiplicity: int = 1,
         comment_symbols: str | Sequence[str] | None = None,
-        struc_limit: int | None = None,
+        n_struc_limit: int | None = None,
     ) -> "list[Structure]":
         """
         Function for reading a xyz trajectory data string and converting it to a list of molecular Structure
```
```suggestion
        Function for reading a XYZ trajectory data string and converting it to a list of molecular Structure
```

Please consistent with casing of names and terms. At least be consistent with in a module.
But, XYZ should be spelled all-caps. Please adapt every occurrence.

src/opi/input/structures/structure.py
```diff
@@ -1026,3 +995,29 @@ def from_lists(
...
+    @classmethod
+    def _iter_xyz_structures(
+        cls,
+        tracked: TrackingTextIO,
+        charge: int = 0,
+        multiplicity: int = 1,
+        comment_symbols: str | Sequence[str] | None = None,
+        n_struc_limit: int | None = None,
+    ) -> Iterator["Structure"]:
+        """Yield properties from the buffer until exhausted or the limit is reached."""
```
timmyte: Please insert a complete docstring.

src/opi/input/structures/properties.py
```diff
@@ -0,0 +1,322 @@
...
+        numbers = [float(m) for m in re.findall(RGX_INT_AND_FLOAT, line)]
+        # > Parse the comment line
+        try:
+            energy_total = numbers[0]
+        except (IndexError, ValueError) as err:
+            raise ValueError("Could not parse goat energies from comment line.") from err
```
timmyte:
```suggestion
            raise ValueError("Could not parse GOAT energies from comment line.") from err
```

examples/README.md
```diff
@@ -70,4 +70,5 @@ python3 job.py
 - exmp047_trj: Perform calculations on all structures of a multi-xyz (trj) file
 - exmp048_loc_gbw: Read localized MO coefficients and transform integrals in LMO basis
 - exmp049_arbitrary_block_variable: Use arbitrary block variable in the SCF block
+- exmp050_docker: Use docker to dock water with water
```
timmyte:
```suggestion
- exmp050_docker: Use DOCKER to dock water with water
```

src/opi/input/structures/properties.py
```diff
@@ -0,0 +1,322 @@
...
+        # > Parse the comment line
+        try:
+            # > ID of the structure
+            structure_id = int(numbers[0])
+            # > Total energy is second number (Eh)
+            energy_total = numbers[1]
+            # > Relative energy is the third number (kcal/mol)
+            energy_relative = numbers[2]
+        except (IndexError, ValueError) as err:
+            raise ValueError("Could not parse docker energies from comment line.") from err
```
timmyte:
```suggestion
            raise ValueError("Could not parse DOCKER energies from comment line.") from err
```

timmyte (changes requested):

src/opi/input/structures/properties.py
haneug: Indeed this is inconsistent and this is also the case with `Structure`. The None return type indicated the end of the buffer and I replaced it now by raising an `EOFError` which should be more consistent. 
In the `_trj` routines I know that we reached the end of the file without reading anything if the list is empty and no `ValueError` has been raised.

PR #155, opened 2025-11-12
added Input.format_before_coords() and Input.format_after_coords()

### LINKED ISSUE #62: Add "Input.format()"
The `Input` class is not self-contained, but relies on `Calculator.write_input()` to format and write the input.
But the class should have its on method `format()` that outputs the input as formatted text.
This should be called by `Calculator.write_input()`

nakul680:
Closes #62 
- Created `Input.format_before_coords()` and `Input.format_after_coords()` . These methods are then called by `Calculator.write_input()`

### DIFF
--- a/src/opi/input/core.py
+++ b/src/opi/input/core.py
@@ -167,14 +167,14 @@ class Input:
     # > METHODS
     # %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
 
-    def format_before_coords(self, working_directory: Path | None) -> str:
+    def format_before_coords(self, working_directory: Path | None = None) -> str:
         """
         Function to format input data that appears before the coordinate block in the ORCA .inp file.
 
         Parameters
         ----------
-        working_directory: Path
-            working directory of the ORCA calculation.
+        working_directory: Path | None = None
+            working directory of the ORCA calculation. If not given , current working directory is used.
 
         Raises
         ------


src/opi/input/core.py
```diff
@@ -167,6 +167,120 @@ def moinp(self, value: Path | str | os.PathLike[str] | None) -> None:
...
+
+        Parameters
+        ----------
+        working_directory: Path
+            working directory of the ORCA calculation.
+
+        Raises
+        ------
+        ValueError
+            If the working directory is empty
```
timmyte: Please add comment to the line where the Exception might be raised.

src/opi/input/core.py
```diff
@@ -167,6 +167,120 @@ def moinp(self, value: Path | str | os.PathLike[str] | None) -> None:
...
+        except ValueError as err:
+            raise ValueError(f"Error formatting Input: {err}") from err
+
+    def format_after_coords(self) -> str:
+        """
+        Function to format input data that appears after the coordinate block in the ORCA .inp file.
+
+        Raises
+        ------
+        ValueError
```
timmyte, line 250: Please add more details. When is this raised?
Can this be raised at every point in the function?

src/opi/input/core.py
```diff
@@ -167,6 +167,120 @@ def moinp(self, value: Path | str | os.PathLike[str] | None) -> None:
...
+
+        """
+        input_after_coords = ""
+        try:
+            blocks = self.blocks.values() if self.blocks else ()
+            arbitrary_strings = self.arbitrary_strings
+            # ---------------------------------
+            # > Block options: After coords
+            # ---------------------------------
+            if blocks:
```
timmyte: The `if` can be omitted, as `blocks` is an empty list, so the for-loop does nothing.

src/opi/input/core.py
```diff
@@ -167,6 +167,120 @@ def moinp(self, value: Path | str | os.PathLike[str] | None) -> None:
...
+                input_before_coords += f"%maxcore {memory:d}\n"
+            if (ncores := self.ncores) is not None:
+                input_before_coords += f"%pal\n    nprocs {ncores:d}\nend\n"
+            if (moinp := self.moinp) is not None:
+                input_before_coords += f'%moinp "{moinp.relative_to(working_directory)}"\n'
+
+            # ---------------------------------
+            # > Block Options: Before coords
+            # ---------------------------------
+            if blocks:
```
timmyte: Please add more details. When is this raised?
Can this be raised at every point in the function?

src/opi/core.py
```diff
@@ -215,55 +214,15 @@ def write_input(self, force: bool = True) -> bool:
 
         try:
             input_param = self.input
-            simple_keywords = input_param.simple_keywords
-            blocks = input_param.blocks.values() if input_param.blocks else ()
-            arbitrary_strings = input_param.arbitrary_strings
 
             assert self.inpfile is not None
-            with self.inpfile.open("w") as inp:
+            with open(self._inpfile, "w") as inp:
```
timmyte: Why would you change the `open()` method?

src/opi/core.py
```diff
@@ -215,55 +214,15 @@ def write_input(self, force: bool = True) -> bool:
...
                 # ---------------------------------
-                # > Arbitrary Strings: top
+                # > Before coords block
                 # ---------------------------------
-                if arbitrary_strings:
-                    for item in arbitrary_strings:
-                        if item.pos is ArbitraryStringPos.TOP:
-                            inp.write(f"{item.format_orca()}\n")
+                input_string_before_coords = input_param.format_before_coords(self.working_dir)
```
timmyte:
```suggestion
                inp.write(
                    input_param.format_before_coords(
                        self.working_dir
                ))
```

src/opi/core.py
```diff
@@ -275,20 +234,11 @@ def write_input(self, force: bool = True) -> bool:
...
 
                 # ---------------------------------
-                # > Block options: After coords
+                # > After coords block
                 # ---------------------------------
-                if blocks:
-                    for block in blocks:
-                        if block.aftercoord:
-                            inp.write(f"\n{block.format_orca()}\n")
+                input_string_after_coords = input_param.format_after_coords()
```
timmyte:
```suggestion
                inp.write(
                    input_param.format_after_coords()
                )
```

src/opi/input/core.py
nakul680: Could you clarify this point a bit more?

timmyte: Sorry.
Looks like I placed the comment on the wrong line.
It's meant to apply to this part here:
```
        Raises
        ------
        ValueError
            If the working directory is empty
```

src/opi/input/core.py
```diff
@@ -167,6 +167,119 @@ def moinp(self, value: Path | str | os.PathLike[str] | None) -> None:
     # > METHODS
     # %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
 
+    def format_before_coords(self, working_directory: Path) -> str:
```
timmyte: Please make `working_directory` optional.
If it's not set, fallback to `Path.cwd()`.

BTW: `working_directory: Path = Path.cwd()` is not the way to this.

src/opi/input/core.py
```diff
@@ -167,6 +167,120 @@ def moinp(self, value: Path | str | os.PathLike[str] | None) -> None:
     # > METHODS
     # %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
 
+    def format_before_coords(self, working_directory: Path | None) -> str:
```
timmyte: `working_directory` is still not optional.
Please make it optional and update the docstring.

timmyte (approved):
Looks good. Nice work!

PR #159, opened 2025-11-19
Create `PropertyResults` and `GbwResults` objects from json files 

### LINKED ISSUE #66: `PropertyResults` and `GbwResults` without `Output`
We should make sure that  `PropertyResults` and `GbwResults` also work without `Output`, i.e. by having a class method `from_json()`.
* But this is not the officially recommended way.

nakul680:
- added `from_json()` class method to `PropertyResults` and `GbwResults` class.
- closes #66


src/opi/output/models/json/property/property_results.py
```diff
@@ -31,3 +35,38 @@ class PropertyResults(GetItem):
...
+        try:
+            with open(json_file, "r") as file:
+                data = json.load(file)
+        except FileNotFoundError:
+            raise FileNotFoundError(f"File {json_file} not found")
+
+        data = dict_to_lower(data)
+
+        if not isinstance(data, dict):
+            raise AssertionError("Data is not a dictionary")
```
timmyte: One does not raise `AssertionErrors` manually.
This would be `TypeError`.

src/opi/utils/dict_to_lower.py
```diff
@@ -0,0 +1,26 @@
...
+    Recursively convert all key values in a dictionary to lowercase.
+
+    Parameters
+    ----------
+    obj: dict
+        Dictionary to convert to lowercase.
+
+    Returns
+    -------
+    dict
```
timmyte: Are you sure that only a dict can returned?
What's is returned in the `else` block?

src/opi/utils/dict_to_lower.py
```diff
@@ -0,0 +1,26 @@
...
+        Dictionary converted to lowercase.
+
+    """
+    if isinstance(obj, dict):
```
timmyte: The type hint in the function header says: `dict[str, Any]`, while in this if-else tree below you also check other types?
Which part is wrong?

src/opi/utils/dict_to_lower.py
```diff
@@ -0,0 +1,26 @@
...
+def dict_to_lower(
+    obj: dict[str, Any],
+) -> dict[str, Any] | list[dict[str, Any]]:
+    """
+    Recursively convert all key values in a dictionary to lowercase.
+
+    Parameters
+    ----------
+    obj: dict
+        Dictionary to convert to lowercase.
```
timmyte: Only dictionaries?

src/opi/output/models/json/property/property_results.py
```diff
@@ -31,3 +35,38 @@ class PropertyResults(GetItem):
...
+        ------
+        AssertionError
+            Raised if `dict_to_lower()` does not return a dictionary.
+        FileNotFoundError
+            Raised if `json_file` does not point to a file.
+
+        """
+        try:
+            with open(json_file, "r") as file:
+                data = json.load(file)
```
timmyte: Can `json.load()` raise any exceptions? If yes, which ones?

src/opi/output/models/json/gbw/gbw_results.py
```diff
@@ -26,3 +30,36 @@ class GbwResults(GetItem):
 
     class Configuration:
         allow_population_by_field_name = True
+
+    @classmethod
+    def from_json(cls, json_file: Path | str) -> "GbwResults":
```
timmyte: 1) Please outsource this method into a separate class and let these classes inherit from it. Similarly to how we did it with `GetItem`.
2) Careful. IMO it's more flexible if `from_json()` would take a JSON string rather than a file path. Thereby the user can decide where the JSON data comes from.
3) Please add a method, `from_json_file()` which wraps the former one.

src/opi/output/models/json_loadable.py
```diff
@@ -0,0 +1,74 @@
...
+from pathlib import Path
+from typing import Any, Self
+
+from opi.output.models.base.get_item import GetItem
+from opi.utils.dict_to_lower import dict_to_lower
+
+
+class JSONLoadable(GetItem):
+    """
+    Superclass providing utility methods for loading objects from JSON data or files.
```
timmyte, line 12: Please turn this class into an abstract class, to emphasize that this class it not be instantiated.
```
class JSONLoadable(GetItem, ABC):
```
Same applies for `GetItem`.

src/opi/output/models/json_loadable.py
```diff
@@ -0,0 +1,74 @@
...
+    """
+
+    @classmethod
+    def from_json(cls, json_data: dict[str, Any]) -> Self:
+        """
+        Loads object from JSON data
+        Parameters
```
timmyte, line 20: Add a blank line above please.

src/opi/output/models/jsonreadable.py
```diff
@@ -0,0 +1,74 @@
...
+
+class JSONLoadable(GetItem):
+    """
+    Superclass providing utility methods for loading objects from JSON data or files.
+    """
+
+    @classmethod
+    def from_json(cls, json_data: dict[str, Any]) -> Self:
+        """
+        Loads object from JSON data
```
timmyte:
```suggestion
        Create object from JSON data
```

src/opi/output/models/json_loadable.py
```diff
@@ -0,0 +1,74 @@
...
+        if not isinstance(data, dict):
+            raise TypeError("Data is not a dictionary")
+
+        return cls(**data)
+
+    @classmethod
+    def from_json_file(cls, json_file: Path | str) -> Self:
+        """
+        Creates an object from a JSON file.
+        Parameters
```
timmyte, line 47:
```suggestion

        Parameters
```

src/opi/output/models/jsonreadable.py
```diff
@@ -0,0 +1,74 @@
...
+        """
+        if isinstance(json_file, str):
+            json_file = Path(json_file)
+        try:
+            with open(json_file, "r") as file:
+                data = json.load(file)
+        except FileNotFoundError:
+            raise FileNotFoundError(f"File {json_file} not found")
+        except json.decoder.JSONDecodeError:
+            raise ValueError("Invalid JSON")
```
timmyte: Please add the original source (file path) to the exception message.

src/opi/utils/dict_to_lower.py
```diff
@@ -0,0 +1,29 @@
+from typing import Any
+
+
+def dict_to_lower(obj: dict[str, Any] | list[Any] | Any) -> dict[str, Any] | list[Any] | Any:
```
timmyte: Please put that into `misc`.
'bit overkill to have that in separate module.

src/opi/output/models/json/gbw/gbw_results.py
```diff
@@ -1,12 +1,12 @@
 from pydantic import Field
 
-from opi.output.models.base.get_item import GetItem
 from opi.output.models.json.gbw.properties.cite import Cite
 from opi.output.models.json.gbw.properties.header import OrcaHeader
 from opi.output.models.json.gbw.properties.molecule import Molecule
+from opi.output.models.jsonreadable import JSONLoadable
```
timmyte:
```suggestion
from opi.output.models.json_loadable import JSONLoadable
```

Please pick one name here: either the module and the class both should use the term "loadable" or "readable". But don't mix them!

PR #170, opened 2026-01-09
WIP: Outputunittest

nakul680:
- added output side unit tests for getters 
- added attribute coverage tests for JSONS

### DIFF
--- a/src/opi/input/structures/coordinates.py
+++ b/src/opi/input/structures/coordinates.py
@@ -48,12 +48,12 @@ class Coordinates:
 
         self._coordinates = tup3d
 
-    def to_list(self) -> list[np.float64]:
+    def to_list(self) -> list[float]:
         """
         Returns coordinates as list
         """
         coords = self.coordinates.tolist()
-        return cast(list[np.float64], coords)
+        return cast(list[float], coords)
 
     @property
     def x(self) -> np.float64:


src/opi/input/structures/coordinates.py
```diff
@@ -52,7 +52,7 @@ def to_list(self) -> list[float] | tuple[()]:
         """
         Returns coordinates as list
         """
-        if self.coordinates:
+        if all(self.coordinates):
```
haneug: If I am not mistaken `all` checks all elements for truthiness. So this would fail if I have a single atom in the origin.
Why did you make that change?

src/opi/output/models/base/get_item.py
```diff
@@ -113,3 +113,8 @@ def graph(self, depth: int = -1, level: int = 0, /, *, max_list_length: int = 5)
                 lines.append(header)
 
         return "\n".join(lines)
+
+    def __eq__(self, other: Any) -> bool:
+        if type(self) is not type(other):
+            return False
+        return bool(self.__dict__ == other.__dict__)
```
haneug, line 123: Why dict you wrap this in bool? Is it necessary here? I think not!

src/opi/output/models/json/property/properties/geometries.py
```diff
@@ -129,9 +126,7 @@ class Geometries(GetItem):
         Contains the information about the SCF calculated quadruple moment
     cipsi_energies: list[CiPsi] | None, default = None
         Contains the information about the CIPSI calculation
-    energy_extrapolation: EnergyExtrapolation | None, default = None
-        Contains information about the energy extrapolation
-    roci_energy: list[RoCiEnergy] | None, default = None
+    rocis_energies: list[RoCiEnergy] | None, default = None
```
haneug, line 129: please also update the class name to RoCis, also it is not a list anymore!

src/opi/output/models/json/property/property_results.py
```diff
@@ -6,7 +6,6 @@
     CalculationTiming,
 )
 from opi.output.models.json.property.properties.geometries import Geometries
-from opi.output.models.json.property.properties.pal import PalFlags
 from opi.output.models.json_loadable import JSONLoadable
```
haneug, line 9: please remove pal.py if it is not needed anymore

tests/fixtures/json_files/bs.json
```diff
@@ -0,0 +1,2085 @@
+{
```
haneug: fixtures/json_files should be deleted

tests/unit/test_input_structure_from_other.py
```diff
@@ -0,0 +1,68 @@
+# import pytest
```
haneug: Why did you add this commented file?

tests/unit/test_output_attributes.py
```diff
@@ -0,0 +1,256 @@
+import inspect
```
haneug, line 1: I find it difficult to read this file. Please do the following:
1. Put the content of this file into a `Test...` class and put the actual test at the beginning of this class
2. Put all helper functions below the test function
3. Document all functions with numpy style doc strings as we normally do. It should be directly clear to anyone reading this file what is actually tested and what the task of certain helper functions is! Why do we even need JSON files here? This is not obvious for someone new.
4. do not use `tempfile` but [tmp_path](https://docs.pytest.org/en/stable/how-to/tmp_path.html) to let pytest handle the clean up. Also do you even need to copy the JSON files? I think `Output` can just read them in place?

tests/unit/test_output_get_values.py
```diff
@@ -0,0 +1,1487 @@
+import glob
```
haneug: We do not want to use the fixtures anymore. I see that we are missing the ORCA outputs. We should probably discuss this in person.

tests/unit/test_output_attributes.py
```diff
@@ -0,0 +1,256 @@
...
+from typing import Set, get_args, get_origin
+
+import pytest
+from pydantic import BaseModel
+
+from opi.output.core import Output
+from opi.output.models.json.gbw.gbw_results import GbwResults
+from opi.output.models.json.property.property_results import PropertyResults
+
+JSON_DIR = Path(__file__).parent.parent / "json_files"
```
haneug: JSON_DIR should be included in the `conftest.py` file. I would think about something like this:
```
JSON_DIR = Path(__file__) / "json_files"

@pytest.fixture(scope="session")
def json_dir() -> Path:
    """
    Path to the JSON directory.
    """
    return JSON_DIR
``` 
and then you can use `Path: json_files` as argument in your tests

tests/unit/test_output_get_values.py
haneug: We discussed that the tests for the getters should be separated into multiple files, and we leave out any getters that rely on secondary binaries or the output file. The output file ones could be added in a separate PR.

src/opi/output/models/base/get_item.py
nakul680: When it is not explicitly casted to bool this statement raises a mypy error. I believe it is to do with the fact that other is of type Any

src/opi/input/structures/coordinates.py
```diff
@@ -52,7 +52,7 @@ def to_list(self) -> list[float] | tuple[()]:
         """
         Returns coordinates as list
         """
-        if self.coordinates:
+        if self.coordinates.size > 0:
```
haneug: I am still wondering why an empty tuple is returned and not an empty list. @timmyte is there a reason for this?

tests/fixtures/output_object.py
```diff
@@ -0,0 +1,33 @@
+import pytest
+
+from opi.output.core import Output
+
+
+@pytest.fixture
+def output_object_factory(json_dir) -> Output:
+    def _create_instance(identifier):
+        matching_files = list(json_dir.rglob(f"*{identifier}*.json"))
```
haneug: this is quite error prone. Could we have one central list in a fixture with all the json file names? Alternatively one with property and one with gbw json files?
Something like:
```
@pytest.fixture
def json_list(json_dir: Path):
    files = [
        "some.property.json",
        ...
    ]
    return [json_dir / f for f in files]
```
This should be used instead of `rglob` everywhere

tests/unit/test_output_attributes.py
```diff
@@ -0,0 +1,298 @@
...
+        json_files = list(json_dir.rglob(f"{basename}.*"))
+
+        # Separate GBW json and property json
+        gbw_json = next(
+            f for f in json_files if f.suffix == ".json" and not f.name.endswith(".property.json")
+        )
+        property_json = next(f for f in json_files if f.name.endswith(".property.json"))
+
+        # Use pytest-provided temp directory
+        temp_dir = tmp_path
```
haneug: Why is tmp_path required here? Can you not read the json files directly even if they lie in different dirs?

tests/unit/test_output_attributes.py
```diff
@@ -0,0 +1,298 @@
...
+import pytest
+from pydantic import BaseModel
+
+from opi.output.core import Output
+from opi.output.models.json.gbw.gbw_results import GbwResults
+from opi.output.models.json.property.property_results import PropertyResults
+
+
+class TestOutputAttributes:
+    @pytest.mark.unit
```
haneug, line 30: Please add a class docstring that explains this class -> test attributes with helper functions

tests/unit/test_output_attributes.py
```diff
@@ -0,0 +1,298 @@
...
+                        result |= self.collect_non_none_attrs(
+                            item,
+                            depth=depth - 1 if depth > 0 else -1,
+                            prefix=path,
+                            _visited=_visited,
+                        )
+
+        return result
+
+    def make_output_object(self, basename: str, tmp_path: Path, json_dir: Path):
```
haneug: Return type?

tests/unit/test_output_attributes.py
```diff
@@ -0,0 +1,298 @@
...
+        return attributes
+
+    def collect_non_none_attrs(
+        self,
+        obj: Any,
+        *,
+        depth: int = -1,
+        prefix: str | None = None,
+        _visited=None,
+    ):
```
haneug: Return type?

tests/fixtures/output_object.py
```diff
@@ -0,0 +1,33 @@
+import pytest
+
+from opi.output.core import Output
+
+
+@pytest.fixture
+def output_object_factory(json_dir) -> Output:
```
haneug: The return type should be `Callable[[str], Output]` and not Output

tests/fixtures/output_object.py
```diff
@@ -0,0 +1,33 @@
...
+    def _create_instance(identifier):
```
haneug: add type of identifier: 
```
(identifier: str) -> Output:
```

tests/unit/test_output_get_electric_properties.py
```diff
@@ -0,0 +1,104 @@
+import pytest
+
+from opi.output.models.json.property.properties.dipole_moment import DipoleMoment
+from opi.output.models.json.property.properties.polarizability import Polarizability
+from opi.output.models.json.property.properties.quadrupole_moment import QuadrupoleMoment
+
+
```
haneug, line 16: Please add a comment in each output test file what is tested here

tests/unit/test_output_attributes.py
nakul680: i don't think they can

src/opi/input/structures/coordinates.py
timmyte: @haneug 
No. This is a conceptual questions. It depends on the purpose of the method. But it could be adjusted here.

src/opi/output/models/base/get_item.py
timmyte: This lacks a comment describing why we `__eq__` is defined like this.

timmyte (approved):

PR #172, opened 2026-01-14
Update Output

### LINKED ISSUE #171: Update Output parse function for all available file combinations
There are different scenarios for available gbw, property.txt, and json files which should all be considered in `Output`.

haneug:
- Separate parse into `parse_gbw` and `parse_property`
- Change `property_json_file` and `gbw_json_files` to `@property` and allow custom names for the json files via the setter   

closes #171

### DIFF
--- a/src/opi/output/core.py
+++ b/src/opi/output/core.py
@@ -4,7 +4,7 @@ It's mostly based on the ORCA's two JSONs files.
 """
 
 import json
-from collections.abc import Iterable, Sequence
+from collections.abc import Sequence
 from pathlib import Path
 from typing import Any, Callable, cast
 from warnings import warn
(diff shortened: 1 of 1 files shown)


src/opi/output/core.py
```diff
@@ -184,40 +182,100 @@ def parse(
...
+            self.create_missing_gbw_json()
+        elif do_create_gbw_json:
+            self.create_gbw_json(force=True)
+
+        # // read the GBW files
+        self.gbw_json_data = [self._process_json_file(file) for file in self.gbw_json_files]
+        self.results_gbw = [GbwResults(**data) for data in self.gbw_json_data]
+
+    @property
+    def gbw_json_files(self) -> Sequence[Path]:
```
```suggestion
    def gbw_json_files(self) -> tuple[Path, ...]:
```

src/opi/output/core.py
```diff
@@ -184,40 +182,100 @@ def parse(
...
+        return self._gbw_json_files
+
+    @gbw_json_files.setter
+    def gbw_json_files(self, value: Iterable[str | Path]) -> None:
```
timmyte: * About the `Iterable[str | Path]`, I see where you are coming from, but typically `value` will either be a list or a tuple and these are usually summarized under `Sequence[str | Path]`, even though there are also iterable.
* You have to catch `value = None` here, otherwise you will get an error as you cannot enumerate over `None`. I would suggest to just check for it and set `gbw_json_files` to an empty tuple and leave.
* `None` should also be reflected in the type hint.

src/opi/output/core.py
```diff
@@ -184,40 +182,100 @@ def parse(
...
+        if not self.gbw_json_files:
+            self.collect_gbw_json_files()
+        # // Create missing GBW JSON files
+        if do_create_gbw_json is None:
+            self.create_missing_gbw_json()
+        elif do_create_gbw_json:
+            self.create_gbw_json(force=True)
+
+        # // read the GBW files
+        self.gbw_json_data = [self._process_json_file(file) for file in self.gbw_json_files]
```
timmyte: Just a side note:
One think to keep in mind here, is that if processing of even one JSON file fails, the whole chain breaks apart. This is not ideal, as the interface should be able to handle such small potholes.

haneug: That is a good point! I handled this by adding the method `_process_json_files` which should catch the errors raised by `_process.json_file` and continue if desired.

timmyte (approved):
Nice work ;)

PR #173, opened 2026-01-16
extension to method block

nakul680:
- added requested attributes to method block


src/opi/input/blocks/base.py
```diff
@@ -189,3 +189,11 @@ def init_inputpath(cls, inp: Any) -> Any:
             return InputFilePath(file=inp)
         else:
             return inp
+
+    @field_validator("*", mode="before")
```
haneug: You apply this field_validator to everything which results in every string value in every field being lower case. This results in failures for the example tests 16, 18, and 20. I would suggest to put this in the derived class.

examples/README.md
```diff
@@ -71,4 +71,5 @@ python3 job.py
 - exmp048_loc_gbw: Read localized MO coefficients and transform integrals in LMO basis
 - exmp049_arbitrary_block_variable: Use arbitrary block variable in the SCF block
 - exmp050_docker: Use DOCKER to dock water with water
+- exmp051_libxc: Shows modification of DFT LibXC parameters via the method block
```
haneug, line 74: nice!

tests/examples/test_exmp051_libxc.py
```diff
@@ -0,0 +1,18 @@
+import pytest
+
+from examples.exmp051_libxc.job import run_exmp051
+from opi.input.structures import Structure
+
+
+@pytest.mark.examples
+@pytest.mark.orca
+@pytest.mark.slow
+def test_exmp051_libxc(example_input_file, tmp_path) -> None:
```
haneug, line 10: Please add a docstring on what is tested -> successful calculation

src/opi/input/blocks/block_method.py
```diff
@@ -33,3 +135,29 @@ def path_from_string(cls, path: str | InputFilePath) -> InputFilePath:
...
         else:
             return path
+
+    @field_validator("extparamx", "extparamc", "extparamxc", mode="before")
+    @classmethod
+    def init_ext_param_from_string(
+        cls, inp: dict[str, float] | ExternalParam, info: ValidationInfo
+    ) -> ExternalParam:
+        """
+
```
haneug, line 154: A short description what this does would be nice

src/opi/input/blocks/base.py
nakul680: dont we want everything to be lowercase anyway since that is how it is formatted into the inp file?

haneug: It does not fail in ORCA, but the validation for the enums used in these examples fails so no input is written at all.

src/opi/input/blocks/block_method.py
```diff
@@ -1,16 +1,118 @@
...
+        for string in strings:
+            paramname, value = string.rsplit(" ", maxsplit=1)
+            params[paramname.strip()] = float(value.strip())
+
+        return ExternalParam(field_name=field_name, parameters=params)
+
+    def __str__(self) -> str:
+        lines = []
+
+        for i, (key, value) in enumerate(self.parameters.items()):
```
timmyte, line 46: Could we add a bit more description here please, why we have special behavior for the first line.

src/opi/input/blocks/block_method.py
```diff
@@ -1,16 +1,118 @@
...
+    parameters: dict[str, float]
+        Dictionary of parameter values.
+
+    """
+
+    field_name: str | None = None
+    parameters: dict[str, float]
+
+    @classmethod
+    def from_string(cls, field_name: str, strings: list[str]) -> "ExternalParam":
```
timmyte: I would suggest just adding a little example in the docstring, e.g.
```
"""
Create an `ExternalParam` object from string such as: `ExtParamX   "_bt"     0.00444` [...]
"""
```

src/opi/input/blocks/block_method.py
```diff
@@ -1,16 +1,118 @@
...
+        params = {}
+        for string in strings:
+            paramname, value = string.rsplit(" ", maxsplit=1)
```
timmyte, line 38: 1) I would also add a bit more description here about `rsplit(" ", maxsplit=1)`.
2) To me it also seems as if you are storing the quotes around the parameter name in `paramname`. I would strip these as well.

src/opi/input/blocks/block_method.py
```diff
@@ -33,3 +135,39 @@ def path_from_string(cls, path: str | InputFilePath) -> InputFilePath:
...
+        if isinstance(inp, ExternalParam):
+            if inp.field_name is None:
+                inp.field_name = field_name
+            return inp
+        else:
+            return ExternalParam(field_name=field_name, parameters=inp)
+
+    @field_validator("*", mode="before")
+    @classmethod
+    def string_tolower(cls, inp: Any) -> Any:
```
timmyte: I would also add a docstring here. What exactly is converted to lowercase?
This would also clarify the necessity of if-statement below.

src/opi/input/blocks/block_method.py
nakul680: I designed such that the user does not have to put in the quotes when they give the input, as if OPI is adding the quotes internally when formatting i dont see any reason for the user to have to add quotes as well. Should I change this ?

timmyte: No. Thanks for the explanation. I just misunderstood the code here.

PR #190, opened 2026-02-05
Create IntgroupEnd class and fix bug in BlockQmmm

### LINKED ISSUE #187: QMAtoms List Not Written Correctly
Some integer lists in orca require an `end` keyword at the end, some examples are `qmatoms` and `activeatoms` in the qmmm input block as described in the [documentation](https://www.faccts.de/docs/orca/6.1/manual/contents/multiscalesimulations/qmmm-general.html?q=active+region&n=1#qm-and-active-region-definition).

**To Reproduce**
```python
>>> from opi.input.blocks import BlockQmmm
>>> block = BlockQmmm(qmatoms=[1, 2, 3, 4])
>>> print(block.format_orca())
%qmmm
    qmatoms { 1 2 3 4 }
end
``` 

**Expected behavior**
The above example should output the following:
```
%qmmm
    qmatoms { 1 2 3 4 } end
end
```

I'm not sure where else this is an issue as they should also be identified. Depending on how you want to structure the code a simple fix is to replace the type of `IntGroup`  in this example with something like `IntGroupEnd`. An example implementation would be as follows:

```python
class IntGroupEnd(IntGroup):
    def __str__(self):
        return f"{super().__str__()} end"
```

Then update the types and field validators where applicable. Let me know if this approach suits the ethos of the repo or you would like a completely separate class for handling this type of input.

nakul680:
- created class IntGroupEnd
- move IntGroupEnd and IntGroup to models folder 
- fix #187  

Refactoring code will be done in separate PR

### DIFF
--- a/src/opi/input/blocks/block_qmmm.py
+++ b/src/opi/input/blocks/block_qmmm.py
@@ -132,7 +132,7 @@ class BlockQmmm(Block):
         mode="before",
     )
     @classmethod
-    def intgroup_from_list(cls, inp: IntGroupEnd | list[int]) -> IntGroupEnd:
+    def intgroupend_from_list(cls, inp: IntGroupEnd | list[int]) -> IntGroupEnd:
         """
         Parameters
         ----------


src/opi/input/blocks/block_qmmm.py
```diff
@@ -131,14 +132,14 @@ class BlockQmmm(Block):
         mode="before",
     )
     @classmethod
-    def intgroup_from_list(cls, inp: IntGroup | list[int]) -> IntGroup:
+    def intgroup_from_list(cls, inp: IntGroupEnd | list[int]) -> IntGroupEnd:
```
haneug: `intgroupend_from_list`

PR #197, opened 2026-02-18
Release/2.0 Fixes into main (#196)

haneug:
- Adds README for the testsuite
- Update noxfile so that `-s tests` now only runs the unit tests
- Fixes bug in `Output.config_dict` and adapts the Fock notebook 
- Updates the `CHANGELOG.md` to improve the structure and overview
- Update main README so that stable points to the new 2.0 documentation

### DIFF
--- a/src/opi/output/core.py
+++ b/src/opi/output/core.py
@@ -2023,8 +2023,10 @@ class Output:
         if recreate_json:
             if self.config_dict is None:
                 self.config_dict = {}
+            # // 1elIntegrals
             if "1elIntegrals" not in self.config_dict:
                 self.config_dict["1elIntegrals"] = []
+            # // S Integrals
             if "S" not in self.config_dict["1elIntegrals"]:
                 self.config_dict["1elIntegrals"].append("S")
             self.recreate_gbw_results(self.config_dict, gbw_index)
@@ -2056,8 +2058,10 @@ class Output:
         if recreate_json:
             if self.config_dict is None:
                 self.config_dict = {}
+            # // 1elIntegrals
             if "1elIntegrals" not in self.config_dict:
                 self.config_dict["1elIntegrals"] = []
+            # // H Integrals
             if "H" not in self.config_dict["1elIntegrals"]:
                 self.config_dict["1elIntegrals"].append("H")
             self.recreate_gbw_results(self.config_dict, gbw_index)
@@ -2088,13 +2092,14 @@ class Output:
         if recreate_json:
             if self.config_dict is None:
                 self.config_dict = {}
+            # // FockMatrix
             if "FockMatrix" not in self.config_dict:
                 self.config_dict["FockMatrix"] = []
+            # // F (full Fock matrix)
             if "F" not in self.config_dict["FockMatrix"]:
                 self.config_dict["FockMatrix"].append("F")
             self.recreate_gbw_results(self.config_dict, gbw_index)
 
-        # > get hcore from gbw json files
         fock_list = self._safe_get("results_gbw", gbw_index, "molecule", "f_matrix")
 
         if fock_list is not None:
@@ -2120,13 +2125,14 @@ class Output:
         if recreate_json:
             if self.config_dict is None:
                 self.config_dict = {}
+            # // FockMatrix
             if "FockMatrix" not in self.config_dict:
                 self.config_dict["FockMatrix"] = []
+            # // J contribution to Fock matrix
             if "J" not in self.config_dict["FockMatrix"]:
                 self.config_dict["FockMatrix"].append("J")
             self.recreate_gbw_results(self.config_dict, gbw_index)
 
-        # > get hcore from gbw json files
         j_list = self._safe_get("results_gbw", gbw_index, "molecule", "j_matrix")
 
         if j_list is not None:
@@ -2152,13 +2158,14 @@ class Output:
         if recreate_json:
             if self.config_dict is None:
                 self.config_dict = {}
+            # // FockMatrix
             if "FockMatrix" not in self.config_dict:
                 self.config_dict["FockMatrix"] = []
+            # // K contribution to Fock matrix
             if "K" not in self.config_dict["FockMatrix"]:
                 self.config_dict["FockMatrix"].append("K")
             self.recreate_gbw_results(self.config_dict, gbw_index)
 
-        # > get hcore from gbw json files
         k_list = self._safe_get("results_gbw", gbw_index, "molecule", "k_matrix")
 
         if k_list is not None:
--- a/tests/README.md
+++ b/tests/README.md
@@ -1,11 +1,7 @@
 # Tests
 
 This directory contains the test suite for **OPI**.\
-Tests are executed via **nox**, typically using **uv**.
-
-By default, only tests that **do not require ORCA** are run.\
-Tests that require ORCA must be requested explicitly and are never run
-in CI.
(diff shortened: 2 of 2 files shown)


tests/README.md
```diff
@@ -0,0 +1,148 @@
...
+
+    uv run nox -s tests
+
+### Default behavior (`-s tests`)
+
+Running
+
+    uv run nox -s tests
+
+executes only tests that **do not require ORCA**.
```
timmyte: Are you sure that this is still true. I was under the impression it runs only units tests now. Therefore `tests` is also not a good name, as it implies that we run the full test framework.
We should also have a Nox session that runs the full testsuite, e.g. `nox -s tests-all/tests-full`

tests/README.md
```diff
@@ -0,0 +1,148 @@
...
+
+    tests/json_files/property
+
+and 
+
+    tests/json_files/gbw
+
+To update these JSON fixtures, run:
+
+    uv run nox -s tests -- -m json_files --update-json-files
```
timmyte: One of the ideas behind using Nox is to have predefined workflows here, which can be invoke with simple commands.
So if we introduce new common workflows these should be added to the noxfile instead of documenting long commands again.

.github/workflows/pr_checks.yml
```diff
@@ -44,4 +44,4 @@ jobs:
       - name: Run Nox sessions via uv
         run: |
           uv run nox -t pr_check
-          uv run nox -s tests -- -m "not orca"
+          uv run nox -s tests
```
timmyte: Add a comment that these are just unit tests.

CHANGELOG.md
```diff
@@ -2,87 +2,87 @@
...
+
+### Added
+#### Input
+- Added support for adding arbitrary options to blocks which will be printed in the block along with the other options. #61
+- Added class methods to initialize `Structure` class: #96
   - `from_ase()` : from Atoms object of ASE
   - `from_list()`: from a list
-  - `from_xyz_string()`: to directly read XYZ string
+  - `from_xyz_block()`: to directly read XYZ string
   - `from_trj_xyz()`: reads multi-XYZ files and returns list of Structure objects.(#138)
```
timmyte:
```suggestion
  - `from_trj_xyz()`: reads multi-XYZ files and returns list of Structure objects. #138
```

timmyte (changes requested):
Thanks for making the final changes. Please check my few comments.

tests/README.md
haneug: I agree `tests` will run all the tests and the unit tests will be put into `unit_tests`

tests/README.md
haneug: Set up a nox session with the name `json_gen` for this.

.github/workflows/pr_checks.yml
haneug: I added a comment but it is now clearer since the session is named unit_tests

tests/README.md
```diff
@@ -0,0 +1,140 @@
...
+
+## Running tests
+
+Tests are run via **nox**. A typical invocation is:
+
+    uv run nox -s tests
+
+### Default behavior (`-s tests`)
+
+this executes all tests.
```
timmyte:
```suggestion
This executes all tests.
```

tests/README.md
```diff
@@ -0,0 +1,140 @@
...
+
+Running
+
+    uv run nox -s unit_tests
+
+Will execute only the unit tests.
```
timmyte:
```suggestion
will execute only the unit tests.
```

tests/README.md
```diff
@@ -0,0 +1,140 @@
+# Tests
+
+This directory contains the test suite for **OPI**.\
+Tests are executed via **nox**, typically using **uv**.
+
+By default, only tests that **do not require ORCA** are run.\
+Tests that require ORCA must be requested explicitly and are never run
+in CI.
```
timmyte:
```suggestion
Tests that require ORCA must be requested explicitly.
```

tests/README.md
```diff
@@ -0,0 +1,140 @@
...
+1.  **Unit tests (CI-safe)**\
+    These tests do **not** require ORCA and are run by default.
+    -   Input-side tests (OPI → ORCA input generation)
+    -   Output-side tests (JSON → Pydantic models → getters)
+    -   General unit tests but no input formatting tests
+2.  **Example tests (local-only)**\
+    These tests execute the examples using ORCA and can therefore only be run locally.
+    -   Used for end-to-end validation
+    -   Also used to generate/update JSON files for output unit tests in the CI
+    -   Must be explicitly requested (marker `examples`)
```
timmyte:
```suggestion
    -   Must be explicitly requested (Pytest marker `examples`)
```

tests/README.md
```diff
@@ -0,0 +1,140 @@
+# Tests
+
+This directory contains the test suite for **OPI**.\
+Tests are executed via **nox**, typically using **uv**.
```
timmyte:
```suggestion
This project uses the Pytest framework.
Tests are typically executed via: `uv run nox ...`, which requires the optional dependency group `tests` to be installed.
```

tests/README.md
```diff
@@ -0,0 +1,140 @@
...
+    -   Output-side tests (JSON → Pydantic models → getters)
+    -   General unit tests but no input formatting tests
+2.  **Example tests (local-only)**\
+    These tests execute the examples using ORCA and can therefore only be run locally.
+    -   Used for end-to-end validation
+    -   Also used to generate/update JSON files for output unit tests in the CI
+    -   Must be explicitly requested (marker `examples`)
+
+Example code lives in the top-level `examples/` directory but is
+**imported and executed by tests** defined under `tests/examples`.
```
timmyte:
```suggestion
Example code lives in the top-level `examples/` directory while the corresponding tests are defined `tests/examples`.
```

tests/README.md
```diff
@@ -0,0 +1,140 @@
...
+Running
+
+    uv run nox -s unit_tests
+
+Will execute only the unit tests.
+This is the mode used in CI.
+
+------------------------------------------------------------------------
+
+## ORCA example tests
```
timmyte: I would remove this section and put all details into corresponding entry in `Test markers`.

tests/README.md
```diff
@@ -0,0 +1,140 @@
...
+
+------------------------------------------------------------------------
+
+## Running tests
+
+Tests are run via **nox**. A typical invocation is:
+
+    uv run nox -s tests
+
+### Default behavior (`-s tests`)
```
timmyte: I would remove the headline and combine the three lines above and the ~5 lines below into one section.

src/opi/output/core.py
```diff
@@ -2025,7 +2025,8 @@ def get_int_overlap(
                 self.config_dict = {}
             if "1elIntegrals" not in self.config_dict:
```
timmyte, line 2027: A brief headline as comment mentioning the property that is checked in the following if statement would make the code a bit more readable.

```suggestion
            # // 1elIntegrals
            if "1elIntegrals" not in self.config_dict:
                ....
            # // S Integrals
            ...
```

This applies too all if-clauses below.

tests/README.md
haneug: This statement is actually not right so I removed it.

tests/README.md
haneug: This statement is actually not right so I removed it

Issue #95, opened 2025-08-08
Simple keywords for LibXC


cplett:
It'd be great to have LibXC support via OPI. Maybe adding respective simple keywords is a good idea.

PR #201, opened 2026-02-20
Libxc

nakul680:
## Closes Issues
-  closes issue #95 

## Description 
- Implementation of simple LibXC keywords


src/opi/input/simple_keywords/libxc_keyword.py
```diff
@@ -0,0 +1,28 @@
+from opi.input.simple_keywords import SimpleKeyword
+
+
+class LibxcKeyword(SimpleKeyword):
```
timmyte: Good idea to implement the LibXC keywords like this.

Could you have a look if you could combine this class with `SolvationModelAndSolvent`?
The underlying behavior is the same: we have a simple keyword that acts like a function that takes an argument. I would suggest introducing a separate parent class for such keywords.
Also to ensure that `__init__()` and `__call__()` are consistently implemented.

haneug: Sorry I failed to mention this here and only talked to Nakul directly about this. The LibXC keywords are actually their own enum. Not every functional is available and some otherwise not available functional should be available with LibXC.

Issue #199, opened 2026-02-18
Improve access to config_dict gbw data


haneug:
It is desirable to improve the overall experience of requesting integrals / density matrices / mo data from the gbw file. 
- [ ] SCF density matrix in AO Basis
- [ ] MO coefficients as numpy arrays

PR #204, opened 2026-02-24
Implement get_scf_density

haneug:
## Closes Issues
#199 - only in part

## Description
* add `scfp`  and `scfr` to the output models and implement a getter function for each.
* orca_2json used the full path -> resulted in the full path being stored in the resulting JSON
* orca_2json did not find any `.densities` files with the complete path
* therefore I changed the call to use only the filename, with the default behavior we execute the command in `working_dir`

---

### Added
* Add SCF (spin-)density matrix in AO basis to output models and implement getter functions `get_scf_density` and `get_scf_spin_density` (#204)

### DIFF
--- a/src/opi/output/core.py
+++ b/src/opi/output/core.py
@@ -1993,7 +1993,8 @@ class Output:
         Parameters
         ----------
         gbw_index: int, default = 0
-            Non-negative index of gbw file in `self.gbw_json_files` for which the json file should be created. Default 0 refers to the main gbw file.
+            Index (>= 0) of the gbw file in `self.gbw_json_files` for which the json file should be created.
+            Negative indices are not allowed. Default 0 refers to the main gbw file.
 
         """
         self.create_gbw_json(force=True, config=config_dict, gbw_index=gbw_index)
@@ -2017,7 +2018,8 @@ class Output:
             If True, recreate the gbw json file and request the overlap integrals to be included.
             The request for these integrals will be added to the `config_dict` attribute.
         gbw_index: int, default = 0
-            Non-negative index of gbw file in `self.gbw_json_files` for which integrals are requested. Default 0 refers to the main gbw file.
+                Index (>= 0) of the gbw file in `self.gbw_json_files` for which integrals are requested.
+                Negative indices are not allowed. Default 0 refers to the main gbw file.
         """
 
         if recreate_json:
@@ -2052,7 +2054,8 @@ class Output:
             If True, recreate the gbw json file and request the hcore integrals to be included.
             The request for these integrals will be added to the `config_dict` attribute.
         gbw_index: int, default = 0
-            Non-negative index of gbw file in `self.gbw_json_files` for which integrals are requested. Default 0 refers to the main gbw file.
+                Index (>= 0) of the gbw file in `self.gbw_json_files` for which integrals are requested.
+                Negative indices are not allowed. Default 0 refers to the main gbw file.
         """
 
         if recreate_json:
@@ -2086,7 +2089,8 @@ class Output:
             If True, recreate the gbw json file and request the fock correction integrals to be included.
             The request for these integrals will be added to the `config_dict` attribute.
         gbw_index: int, default = 0
-            Non-negative index of gbw file in `self.gbw_json_files` for which integrals are requested. Default 0 refers to the main gbw file.
+                Index (>= 0) of the gbw file in `self.gbw_json_files` for which integrals are requested.
+                Negative indices are not allowed. Default 0 refers to the main gbw file.
         """
 
         if recreate_json:
@@ -2119,7 +2123,8 @@ class Output:
             If True, recreate the gbw json file and request J to be included.
             The request for these integrals will be added to the `config_dict` attribute.
         gbw_index: int, default = 0
-            Non-negative index of gbw file in `self.gbw_json_files` for which integrals are requested. Default 0 refers to the main gbw file.
+                Index (>= 0) of the gbw file in `self.gbw_json_files` for which integrals are requested.
+                Negative indices are not allowed. Default 0 refers to the main gbw file.
         """
 
         if recreate_json:
@@ -2152,7 +2157,8 @@ class Output:
             If True, recreate the gbw json file and request K to be included.
             The request for these integrals will be added to the `config_dict` attribute.
         gbw_index: int, default = 0
-            Non-negative index of gbw file in `self.gbw_json_files` for which integrals are requested. Default 0 refers to the main gbw file.
+                Index (>= 0) of the gbw file in `self.gbw_json_files` for which integrals are requested.
+                Negative indices are not allowed. Default 0 refers to the main gbw file.
         """
 
         if recreate_json:
@@ -2185,7 +2191,8 @@ class Output:
             If True, recreate the gbw json file and request scfp to be included.
             The request for the density will be added to the `config_dict` attribute.
         gbw_index: int, default = 0
-            Non-negative index of gbw file in `self.gbw_json_files` for which densities are requested. Default 0 refers to the main gbw file.
+                Index (>= 0) of the gbw file in `self.gbw_json_files` for which densities are requested.
+                Negative indices are not allowed. Default 0 refers to the main gbw file.
         """
         if recreate_json:
             if self.config_dict is None:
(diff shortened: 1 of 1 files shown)


src/opi/execution/core.py
```diff
@@ -192,8 +192,8 @@ def create_gbw_json(
             # > Create JSON-config file if given:
             if config_fmt := self.format_gbw_json_config(config):
                 config_file.write_text(config_fmt)
-            # > Create JSON from GBW file
-            gbw_filename = str(gbw_json_file.with_suffix(suffix))
+            # > Create JSON from GBW file - Use only the filename to not confuse orca
+            gbw_filename = str(gbw_json_file.with_suffix(suffix).name)
             self.run_orca_2json([gbw_filename])
```
timmyte: But if we only use the filename as argument to `orca_2json` I would change this to:

```suggestion
            self.run_orca_2json([gbw_filename], working_dir=gbw_json_file.parent)
```
Not sure if `run_orca_2json()` has `working_dir`. Should be added then.

src/opi/output/models/json/gbw/properties/molecule.py
```diff
@@ -42,6 +43,8 @@ class Molecule(GetItem):
         Pointgroup of the molecule
     td_dft: TdDft | None default = None
         Contains information about td-dft calculation
+    densities: Densities | None default = None
```
timmyte:
```suggestion
    densities: Densities | None, default = None
```

src/opi/output/models/json/gbw/properties/densities.py
```diff
@@ -0,0 +1,16 @@
+from pydantic import StrictFloat
+
+from opi.output.models.base.get_item import GetItem
+
+
+class Densities(GetItem):
```
timmyte, line 6: Don't we have different types of densities, like relaxed and unrelaxed.
Not sure where exactly this information should be stored, as label here in this class, or if different densities should be different properties.
My uneducated preference would probably be the former variant.

src/opi/output/core.py
```diff
@@ -2173,6 +2173,39 @@ def get_int_k(
...
+        """
+        Returns the SCF density matrix.
+
+        Parameters
+        ----------
+        recreate_json : bool, default = False
+            If True, recreate the gbw json file and request K to be included.
+            The request for these integrals will be added to the `config_dict` attribute.
+        gbw_index: int, default = 0
+            Non-negative index of gbw file in `self.gbw_json_files` for which integrals are requested. Default 0 refers to the main gbw file.
```
timmyte: What happens if it's negative?

src/opi/output/core.py
```diff
@@ -2173,6 +2173,39 @@ def get_int_k(
...
+            self.config_dict["Densities"].append("scfp")
+            # // scfp - SCF density matrix P
+            if "scfp" not in self.config_dict["Densities"]:
+                self.config_dict["Densities"].append("scfp")
+            self.recreate_gbw_results(self.config_dict, gbw_index)
+
+        scfp_list = self._safe_get("results_gbw", gbw_index, "molecule", "densities", "scfp")
+
+        if scfp_list is not None:
+            return np.array(scfp_list)
```
```suggestion
            return np.array(scfp_list, dtype=np.float64)
```

src/opi/output/core.py
```diff
@@ -2173,6 +2173,39 @@ def get_int_k(
...
+        ----------
+        recreate_json : bool, default = False
+            If True, recreate the gbw json file and request K to be included.
+            The request for these integrals will be added to the `config_dict` attribute.
+        gbw_index: int, default = 0
+            Non-negative index of gbw file in `self.gbw_json_files` for which integrals are requested. Default 0 refers to the main gbw file.
+        """
+        if recreate_json:
+            if self.config_dict is None:
+                self.config_dict = {}
```
timmyte, line 2199: If we would have a dedidacted class that would represent the config dict for the GBW-JSON we wouldn't always need these "if not exist -> create -> add, otherwise only add" type of if-clauses.
Just sayin'

src/opi/output/models/json/gbw/properties/densities.py
haneug: We are bound to what ORCA provides. I implemented here only the SCF density but there are other densities like e.g. relaxed / unrelaxed MP2 densities and they all should go into this Densities class at some point.

src/opi/output/core.py
haneug: If you request negative numbers recreate will not work and `_safe_get` will just pass the index. So you would either get `None` since nothing is there (recreation silently did nothing) or you get the results for gbw file `[-index]`

src/opi/output/core.py
haneug: I agree, in general the `config_dict` properties should be refactored. Lets discuss that in time

haneug: Added an issue for that #207

examples/README.md
```diff
@@ -72,4 +72,5 @@ python3 job.py
 - exmp049_arbitrary_block_variable: Use arbitrary block variable in the SCF block
 - exmp050_docker: Use DOCKER to dock water with water
 - exmp051_libxc: Shows modification of DFT LibXC parameters via the method block
+- exmp052_densities: Perfor UHF calculation and obtain density and spin-density in AO-basis
```
timmyte:
```suggestion
- exmp052_densities: Perform UHF calculation and obtain density and spin-density in AO-basis
```

src/opi/output/models/json/gbw/properties/densities.py
```diff
@@ -0,0 +1,19 @@
...
+    """
+    This class contains the information about the Densities.
+
+    Attributes
+    ----------
+    scfp: list[list[StrictFloat]]
+        Density matrix from SCF.
+    scfp: list[list[StrictFloat]]
```
timmyte:
```suggestion
    scfr: list[list[StrictFloat]]
```

src/opi/output/models/json/gbw/properties/molecule.py
```diff
@@ -42,6 +43,8 @@ class Molecule(GetItem):
         Pointgroup of the molecule
     td_dft: TdDft | None default = None
         Contains information about td-dft calculation
+    densities: Densities | None, default = None
+        Contains the available densities
```
timmyte, line 47: If possible, I would be more precise which types of densities.
So only SCF densities or also MP2 or maybe other densities.

Just a suggestion -> no showstopper

src/opi/output/core.py
```diff
@@ -2173,6 +2173,70 @@ def get_int_k(
...
+        """
+        Returns the SCF density matrix.
+
+        Parameters
+        ----------
+        recreate_json : bool, default = False
+            If True, recreate the gbw json file and request scfp to be included.
+            The request for the density will be added to the `config_dict` attribute.
+        gbw_index: int, default = 0
+            Non-negative index of gbw file in `self.gbw_json_files` for which densities are requested. Default 0 refers to the main gbw file.
```
timmyte: Mention negative numbers as well, please. What happens if `gbw_index` is negative. Or just state that they are not allowed.

src/opi/output/core.py
```diff
@@ -2173,6 +2173,70 @@ def get_int_k(
...
+        """
+        Returns the SCF spin-density matrix.
+
+        Parameters
+        ----------
+        recreate_json : bool, default = False
+            If True, recreate the gbw json file and request scfr to be included.
+            The request for the density will be added to the `config_dict` attribute.
+        gbw_index: int, default = 0
+            Non-negative index of gbw file in `self.gbw_json_files` for which densities are requested. Default 0 refers to the main gbw file.
```
timmyte: Same here about negative indices.

timmyte (approved):
Thanks for fixing all comments.
PR is good to go. Just fix the last few tiny typos and documentation stuff before merging.

src/opi/output/models/json/gbw/properties/molecule.py
haneug: To my knowledge all densities (will) go in there so for now I would not specify it.

src/opi/output/core.py
haneug: I made it clear that negative indices are not allowed

PR #218, opened 2026-03-16
SimpleTask feature

nakul680:
## Closes Issues
Closes #<issue-id>
...

## Description
- Implements `TaskParams` and `Task` classes
- Implement way to lookup simple keywords from strings 
---

## Release Notes

### Added 

- 

### Changed 

- ...


nakul680:
- Changed `TaskParams` to `Settings`, which have child classes 'TaskSettings` and `MethodSettings`. 
- `Task` class renamed to `SimpleTask`, added `method_settings` and `task_settings` parameters.
- `validate()` function in `Settings` class has been repurposed to contain cross-validation logic (which is not implemented yet) , as `validate_field()` function already handles all validation
- Added documentation 
- created new file `method_settings.py` I think all child classes of `MethodSettings` i.e the 4 major groupings like `DftSettings`, `WftSettings` etc. should live here since they are not task-dependant.

nakul680:
## Latest changes: 
- Introduction of new groupings like `ScfConvergence` , `ScfThreshold` etc. These are now the base classes holding keyword of this type, and the larger grouping `Scf` will inherit from these classes. 
- Fleshed out `DftSettings` class , added cross validation in case of specific methods, and added logic to look for dispersion correction keywords.
- added `strict` flag to avoid overwriting existing working directories. 
- `solvent`, `solvation_model` are now optional keywords. 
-  added aliasing for simple keywords to allow for a better API

examples/exmp054_singlepoint_simpletask/job.py
```diff
@@ -0,0 +1,39 @@
+#!/usr/bin/env python3
+import sys
+from pathlib import Path
+
+from opi.input.structures import Structure
+from opi.simpletasks.singlepointtask import SinglePointResults, SinglePointTask
+
+
+def run_exmp054(
```
haneug: Minor point regarding these examples I would numerate them with increasing complexity:
1. Singlepoint
2. Engrad
3. Opt
4. Freq
5. Goat

src/opi/simple_tasks/tasks/single_point_task.py
```diff
@@ -0,0 +1,64 @@
+import typing
+
+from opi.input.simple_keywords import SimpleKeyword, Solvent, Task
+from opi.simpletasks.base_task import SimpleTask, TaskResults, TaskSettings
+from opi.simpletasks.method_settings import MethodSettings
+
+
+class SinglePointSettings(TaskSettings):
```
haneug, line 7: The Name of this file is not consistent with the class name. It should be `sinlge_point_task.py` to be consisten with `SinglePointTask`

src/opi/simple_tasks/simple_task.py
```diff
@@ -0,0 +1,588 @@
...
+    model_config = ConfigDict(
+        arbitrary_types_allowed=True, validate_assignment=True, extra="forbid"
+    )
+    task_keyword: typing.Annotated[SimpleKeyword, SimpleKeywordBox]
+
+
+_RT = typing.TypeVar("_RT", bound="TaskResults")
+
+
+class SimpleTask(ABC, typing.Generic[_RT]):
```
haneug, line 43: The name of the folder should be consistent with the name of the base class. I would suggest to go with `simple_task` or `simple_tasks`.

src/opi/simple_tasks/tasks/engrad_task.py
```diff
@@ -0,0 +1,80 @@
+import typing
+
+from opi.input.simple_keywords import SimpleKeyword, Solvent, Task
+from opi.simpletasks.base_task import SimpleTask, TaskResults, TaskSettings
+from opi.simpletasks.method_settings import MethodSettings
+
+
+class EngradSettings(TaskSettings):
```
haneug, line 7: I think a sub-folder for the respective Tasks would be appropriate to separate them from the base classes. We expect a lot of simple tasks in the future and while it is easily manageable now I can see that there will be lie 10-20 tasks at some point

examples/exmp059_opt_simpletask/job.py
```diff
@@ -0,0 +1,44 @@
...
+from opi.simpletasks.opt_task import OptResults, OptSettings, OptTask
+
+
+def run_exmp057(structure: Structure | None = None, working_dir: Path = Path("RUN")) -> OptResults:
+    # > if no structure is given read structure from inp.xyz
+    if structure is None:
+        structure = Structure.from_xyz("../exmp054_singlepoint_simpletask/inp.xyz")
+
+    # > set up the task
+    simple_task = OptTask(
```
haneug, line 16: When I delete the method_settings and the task settings from the set up I get the following error:
`AttributeError: 'OptTask' object has no attribute '_task_settings_type'`
The general idea would be that based on the method it is recognized that we need the DFTSettings, right?

haneug (changes requested):
I added some initial comments but I have only scratched the surface so far. I will regularly update my findings here.

examples/exmp059_opt_simpletask/job.py
haneug: And the task settings required should be clear based on OptTask

src/opi/input/simple_keywords/base.py
```diff
@@ -1,8 +1,60 @@
...
 
 
 class SimpleKeywordBox:
-    pass
+    """
+    TODO:
+    - rework registry to account for latest changes.
+    """
+
+    _registry: list[type["SimpleKeywordBox"]] = []
```
haneug: Could you add some explanation regarding the registry and its general idea here?

src/opi/input/simple_keywords/base.py
```diff
@@ -1,8 +1,60 @@
...
+                    return value
+                elif (
+                    isinstance(value, SimpleKeyword) and value.alias and value.alias.lower() == norm
+                ):
+                    return value
+
+        raise ValueError(f"Keyword {s} not found in class {cls.__name__}")
+
+    @classmethod
+    def find_keyword(cls, inp: "SimpleKeyword | str") -> "SimpleKeyword":
```
haneug, line 52: Here I would also appreciate a small docstring

src/opi/input/simple_keywords/base.py
```diff
@@ -1,8 +1,60 @@
...
+    def from_string(cls, s: str) -> "SimpleKeyword":
+        norm = s.lower()
+        for c in cls._registry:
+            for attr in dir(c):
+                if attr.startswith("_"):  # Skip private/magic attributes
+                    continue
+                value = getattr(c, attr)
+                if isinstance(value, SimpleKeyword) and value.keyword.lower() == norm:
+                    return value
+                elif isinstance(value, SimpleKeyword) and attr.lower() == norm:
```
haneug: Could you explain the different cases with some comments?

src/opi/input/simple_keywords/base.py
```diff
@@ -1,8 +1,60 @@
...
+                base._registry.append(cls)
+
+        cls._registry.append(cls)
+
+    @classmethod
+    def registry(cls) -> list[type["SimpleKeywordBox"]]:
+        return cls._registry
+
+    @classmethod
+    def from_string(cls, s: str) -> "SimpleKeyword":
```
haneug, line 13: Here a small docstring explaining what this function does would be helpful for me!

src/opi/simpletasks/base_task.py
```diff
@@ -0,0 +1,588 @@
...
+            raise ValueError("calc_object not set")
+
+        out = self.calc_object.get_output()
+        out.parse()
+        return out
+
+    @cached_property
+    def status(self) -> bool:
+        """``True`` if the job terminated normally and SCF converged."""
+        return self.output.terminated_normally() and self.output.scf_converged()
```
haneug: Here I would just put the `terminated_normally` and not the scf converged since the first one is always present and the second one just for a few methods.

src/opi/simpletasks/engrad_task.py
```diff
@@ -0,0 +1,80 @@
...
+
+    _name: str = "engrad"
+    task_keyword: typing.Annotated[SimpleKeyword, Task] = Task.ENGRAD
+
+
+class EngradResults(TaskResults):
+    """Results from an energy + gradient calculation."""
+
+    @property
+    def final_energy(self) -> float:
```
haneug: Final energy is a property that is actually returned from every calculation so I would even go so far and put it in the base class. Is there a reason why you do not cache this property?

src/opi/simpletasks/engrad_task.py
```diff
@@ -0,0 +1,80 @@
...
+
+    @property
+    def primary_property(self) -> tuple[float, list[float]]:
+        """``(final_energy, gradient)`` tuple."""
+        return self.final_energy, self.gradient
+
+
+class EngradTask(SimpleTask[EngradResults]):
+    """
+    High-level task for single-point energy and gradient calculations.
```
haneug: Here I would remove the High-level from the docstring. I understand how it is meant here but it could result in some confusions when users want a calculation with a "high-level of theory" from ORCA. Therefore, I would just write `Task for xxx`. The same goes for all the other Task docstrings.

src/opi/simple_tasks/simple_task.py
```diff
@@ -0,0 +1,588 @@
...
+        to corresponding `TaskSettings` and `MethodSettings` objects to be configured by user-defined data stored in those
+        objects.
+
+        Returns
+        -------
+        `Input`
+            `Input` object configured by user-defined data.
+
+        """
+        inp = Input()
```
haneug: How is it possible to add simple keywords to the input if the input is recreated here? Is there any additional list for simple keywords the user wants to add to the input?

src/opi/simpletasks/method_settings.py
```diff
@@ -0,0 +1,399 @@
...
+                DispersionCorrection.find_keyword(keywords[1])
+
+                return SimpleKeyword(value)
+            except ValueError:
+                raise ValueError(f"Invalid Dft keyword or dispersion correction given: {value}")
+        else:
+            raise ValueError(f"Invalid Dft keyword '{value}'")
+
+
+class SQMSettings(MethodSettings):
```
haneug: Should be `SqmSettings`

src/opi/simpletasks/method_settings.py
```diff
@@ -0,0 +1,399 @@
...
+            try:
+                if enum_class.find_keyword(method):
+                    return typing.cast(typing.Type[MethodSettings], settings_type)
+            except ValueError:
+                pass
+
+        raise ValueError(f"Keyword {method} not found in any of the valid groupings")
+
+
+class DFTSettings(MethodSettings):
```
haneug: Should be `DftSettings`

src/opi/simpletasks/base_task.py
```diff
@@ -0,0 +1,588 @@
...
+    )
+    task_keyword: typing.Annotated[SimpleKeyword, SimpleKeywordBox]
+
+
+_RT = typing.TypeVar("_RT", bound="TaskResults")
+
+
+class SimpleTask(ABC, typing.Generic[_RT]):
+    """
+    Abstract base class for all high-level OPI calculation tasks.
```
haneug: Here I would also not use `high-level` but rather just use `OPI calculation tasks`

haneug (changes requested):
My question from this round would be: Is it possible to add something to the input object of the Task? Lets say I want to add a specific simple keyword can I do that?

nakul680:
> My question from this round would be: Is it possible to add something to the input object of the Task? Lets say I want to add a specific simple keyword can I do that?

yes i fixed the issue with the input object being newly initialized every time, so now it should be possible

examples/exmp059_opt_simpletask/job.py
nakul680: This error doesnt happen when I delete method_settings and task_settings.  Could you try again with the new changes?

src/opi/simple_tasks/tasks/engrad_task.py
```diff
@@ -0,0 +1,63 @@
+import typing
+
```
haneug, line 2: Here I was a bit unclear I think. I would put all the tasks that one actually uses in one folder. When each task has its own folder one does not get any advantage I think. I would suggest to put them in `tasks`?

haneug:
> > My question from this round would be: Is it possible to add something to the input object of the Task? Lets say I want to add a specific simple keyword can I do that?
> 
> yes i fixed the issue with the input object being newly initialized every time, so now it should be possible

I tried to use it but the input was just empty for me. Also I think using `chached_property` here is not really good style. Can we just have the input as a plain attribute and initialize it in the initialization of task? Or is there a specific reason why you need to have it as a property?

examples/exmp059_opt_simpletask/job.py
haneug: The issue is also resolved for me. Maybe the error was just by me.

src/opi/input/blocks/base.py
```diff
@@ -169,6 +169,12 @@ def format_orca(self) -> str:
 
         return s
 
+    def __add__(self, other: "Block") -> "Block":
+        new_block = self.__class__.model_validate(
```
haneug, line 186: How is it handled if both blocks have the same key with a different value? Please document in the docstring.

src/opi/input/blocks/base.py
```diff
@@ -189,3 +195,16 @@ def init_inputpath(cls, inp: Any) -> Any:
             return InputFilePath(file=inp)
         else:
             return inp
+
+    @classmethod
+    def get_subclass_by_name(cls, name: str) -> type["Block"]:
+        matches = {sub.__name__.lower(): sub for sub in cls.__subclasses__()}
```
haneug: Could you please add some comments here. You set up matches and name_matches, why? You then check matches and name_matches for a match and return it.

src/opi/input/simple_keywords/base.py
```diff
@@ -1,28 +1,124 @@
...
+    across all members in the group.
+    """
+
+    _registry: list[type["SimpleKeywordBox"]] = []
+
+    def __init_subclass__(cls, **kwargs: Any) -> None:
+        super().__init_subclass__(**kwargs)
+        cls._registry = []
+
+        for base in cls.__bases__:
```
haneug: Short description what we do here would be helpful. We check if the base class has a registry and if it has one we add the class to the registry.

src/opi/input/simple_keywords/base.py
```diff
@@ -1,28 +1,124 @@
...
+        attribute name, and the optional alias — in that order.
+
+        Raises
+        ------
+        ValueError
+            If no matching keyword is found in the registry.
+        """
+        norm = s.lower()
+        for c in cls._registry:
+            for attr in dir(c):
```
haneug: A short comment over what we loop here would be helpful I think: We loop over all attributes of the registered classes and then check all the simple keywords from them by skipping private/magic attributes.

haneug: When looking closer at this some things come to my mind: dir will get all the attributes also the inherited class attributes, correct? We could have also used `vars` or `__dict__`. Why is `dir()` used here, this is not tirival (at least to me)! The behavior is what we actually want since the keyword classes are subclasses for the grouping. If we add a function to these classes we would have them here as well in the loop and would not skip them, right? Just something to notice and document or am I mistaken?

src/opi/input/simple_keywords/base.py
```diff
@@ -1,28 +1,124 @@
...
+    """
+    Registry base class for groups of related ``SimpleKeyword`` constants.
+
+    Each concrete subclass (e.g. ``Task``, ``Dft``, ``BasisSet``) declares its
+    members as class-level ``SimpleKeyword`` attributes.  Subclassing
+    automatically registers the new class so that ``from_string`` can search
+    across all members in the group.
+    """
+
+    _registry: list[type["SimpleKeywordBox"]] = []
```
haneug: Please add this attribute to the docstring. I think since this is a list of types this is especially important. What type is it? -> The common base type of all keywords! How is it populated? -> By the init subclass scheme! These points are not obvious!

src/opi/input/simple_keywords/base.py
```diff
@@ -1,28 +1,124 @@
...
+        ------
+        ValueError
+            If no matching keyword is found in the registry.
+        """
+        norm = s.lower()
+        for c in cls._registry:
+            for attr in dir(c):
+                if attr.startswith("_"):  # Skip private/magic attributes
+                    continue
+                value = getattr(c, attr)
```
haneug: I think what was not obvious to me is what value and what attr actually are here. Maybe there are better names for them? like `simple_keyword_from_class` and `simple_keyword_from_class_variable_name`? Being more verbose is worth it here.

src/opi/input/simple_keywords/base.py
```diff
@@ -1,28 +1,124 @@
...
+    A single ORCA simple keyword (e.g. ``SP``, ``PBE``, ``def2-SVP``).
+
+    Instances are used as typed constants inside ``SimpleKeywordBox`` subclasses
+    and can also be constructed directly from a raw string for ad-hoc keywords.
 
     Attributes
     ----------
-    keyword: str
-        simple keyword as it will appear in the ORCA .inp file
+    keyword : str
```
haneug, line 86: We also have name. Do you know where we use that and could you add it to the docstring?

src/opi/simple_tasks/freq/freq_task.py
```diff
@@ -0,0 +1,70 @@
...
+
+    @property
+    def primary_property(self) -> float:
+        """Alias for ``free_energy_delta``."""
+        return float(self.free_energy_delta)
+
+
+class FreqTask(SimpleTask[FreqResults]):
+    """
+    High-level task for harmonic frequency calculations.
```
haneug: High-level

haneug (changes requested):
Nice additions, the input does not work for me yet. Could you add an example? Just request the Hirshfeld charges in the singlepoint via `AtomicCharge.HIRSHFELD`.
I also had some things that I believe need some more explanations in the simplekeyword setup.

src/opi/simple_tasks/tasks/engrad_task.py
nakul680: Ah okay sorry, I misunderstood that, I'll make that change.

src/opi/input/simple_keywords/base.py
nakul680: Thanks for this! This comment made me realize with the current SimpleKeyword structuring that we have right now, we have no need for the registry at all since for example, the parent grouping Scf would already contain the all the attributes from classes like ScfConvergence. In this case, dir() is the right choice since we would also want to see attributes that are inherited from super classes.

src/opi/input/simple_keywords/base.py
nakul680: name was an attribute that should've been removed already. I have removed it now. Thanks!

examples/README.md
```diff
@@ -73,5 +73,9 @@ python3 job.py
 - exmp050_docker: Use DOCKER to dock water with water
 - exmp051_libxc: Shows modification of DFT LibXC parameters via the method block
 - exmp052_densities: Perform UHF calculation and obtain density and spin-density in AO-basis
-- exmp053_opencosmors: Run OpenCOSMO-RS task 
-
+- exmp053_opencosmors: Run OpenCOSMO-RS task
+- exmp054_singlepoint_simpletask: B3LYP/def2-SVP single-point with CPCM(water) using the `SinglePointTask` functionality
+- exmp055_goat_simpletask: GFN2-xTB GOAT conformer search using the `GoatTask` functionality
```
haneug: Please adjust these for the new order.

src/opi/input/blocks/base.py
```diff
@@ -169,6 +169,25 @@ def format_orca(self) -> str:
...
+            Instance of `Block` to be merged into `self`
+
+        Returns
+        -------
+        Block
+            New instance of `Block` with attributes of `self` and `other`.
+
+        """
+        new_block = self.__class__.model_validate(
+            {**self.model_dump(exclude_none=True), **other.model_dump(exclude_unset=True)}
```
haneug: You set up a new block from `self` and `other` and validate the new block. Is there a reason why you use `exclude_none` for self and `exclude_unset`  for other?

src/opi/input/blocks/base.py
```diff
@@ -189,3 +208,42 @@ def init_inputpath(cls, inp: Any) -> Any:
...
+        type[Block]
+            The matching subclass.
+
+        Raises
+        ------
+        ValueError
+            If no subclass matches ``name``.
+        """
+        # Search for `Block` class by OPI name
+        matches = {sub.__name__.lower(): sub for sub in cls.__subclasses__()}
```
haneug: In my opinion the name `matches` is misleading here and also in `name_matches`. First you gather dicts of all the direct subclasses and then you check later if there are any matches. So the variables should rather be `block_name_opi` and `block_name_orca`or similar.

src/opi/input/simple_keywords/base.py
```diff
@@ -1,28 +1,108 @@
...
+
+        Raises
+        ------
+        ValueError
+            If no matching keyword is found in the registry.
+        """
+        norm = s.lower()
+
+        # Here the function will loop over all attributes both existing and inherited by the current class.
+        # The simple keywords are structured such that, for example, the larger `Scf` grouping is a child class of the semantically smaller groupings,
```
haneug, line 28: Here a sentence is missing for me:
`From these attributes we select only the ones which are simple keywords and check if they match the input string.`

src/opi/input/simple_keywords/base.py
```diff
@@ -1,28 +1,108 @@
...
+        # such as `ScfConvergence`, `ScfThreshold` and so on.
+        for attr in dir(cls):
```
haneug: Could we rename `attr` to `attr_name`. It is not the attribute but the name to it?

src/opi/input/simple_keywords/base.py
```diff
@@ -1,28 +1,108 @@
...
+            simple_keyword = getattr(cls, attr)
+            if attr.startswith("_") or not isinstance(
+                simple_keyword, SimpleKeyword
+            ):  # Skip private/magic attributes or non-Simple Keyword attributes
+                continue
+            # Case 1: if the user searches for keyword through how it would appear in ORCA input
+            if simple_keyword.keyword.lower() == norm:
+                return simple_keyword
+            # Case 2: if the user searches for keyword through how it is stored in OPI
+            elif attr.lower() == norm:
```
haneug: since each elif has a return statement you could also just use just if here.

src/opi/input/simple_keywords/base.py
```diff
@@ -1,28 +1,108 @@
...
+        norm = s.lower()
+
+        # Here the function will loop over all attributes both existing and inherited by the current class.
+        # The simple keywords are structured such that, for example, the larger `Scf` grouping is a child class of the semantically smaller groupings,
+        # such as `ScfConvergence`, `ScfThreshold` and so on.
+        for attr in dir(cls):
+            # Get the value associated with attribute
+            simple_keyword = getattr(cls, attr)
+            if attr.startswith("_") or not isinstance(
+                simple_keyword, SimpleKeyword
```
haneug: instead of `simple_keyword` you could go with `candidate` since not all `simple_keyword` will be of instance `SimpleKeyword`

examples/exmp057_singlepoint_simpletask/job.py
```diff
@@ -0,0 +1,43 @@
...
+    # > run the calculation with given data
+    singlepoint_result = simple_task.run("job", structure, working_dir=working_dir)
+
+    # > check if the ORCA calculation terminated normally
+    if not singlepoint_result.status:
+        print("SinglePoint task failed")
+        sys.exit(1)
+
+    # > extract final energy from the `TaskResults` object
+    final_energy = singlepoint_result.final_energy
```
haneug, line 35: I would suggest to also include the `primary_property` in the examples as this is also some form of documentation I think.

src/opi/simple_tasks/simple_task.py
```diff
@@ -0,0 +1,619 @@
...
+    )
+    task_keyword: typing.Annotated[SimpleKeyword, SimpleKeywordBox]
+
+
+_RT = typing.TypeVar("_RT", bound="TaskResults")
+
+
+class SimpleTask(ABC, typing.Generic[_RT]):
+    """
+    Abstract base class for all OPI calculation tasks.
```
haneug: `all OPI calculation tasks` might be too broad. Maybe just go with `for OPI simple tasks.`

src/opi/simple_tasks/tasks/freq_task.py
```diff
@@ -0,0 +1,70 @@
...
+    """Task settings for frequency calculations (``! FREQ``)."""
+
+    _name: str = "freq"
+    task_keyword: typing.Annotated[SimpleKeyword, Task] = Task.FREQ
+
+
+class FreqResults(TaskResults):
+    """Results from a harmonic frequency calculation."""
+
+    @cached_property
```
haneug: I am still not sure when do you cache the properties and when not? e.g. the gradient in the Engrad task is just a property.

src/opi/simple_tasks/settings.py
```diff
@@ -0,0 +1,303 @@
...
+            Tuple of metadata about the field.
+
+        """
+        origin = typing.get_origin(hint)
+        args = typing.get_args(hint)
+        is_union = origin is typing.Union or isinstance(hint, builtin_types.UnionType)
+        if is_union:
+            non_none_args = [arg for arg in args if arg is not type(None)]
+            if non_none_args:
+                return Settings._get_field_metadata(non_none_args[0])
```
haneug, line 146: Here I am a bit puzzled why you use non_none_args[0]. When we do have a Union we just want to look at the first hint? Are there some explicit cases where we have a Union?

src/opi/simple_tasks/settings.py
```diff
@@ -0,0 +1,303 @@
...
+        return args[1:] if len(args) > 1 else ()
+
+    @staticmethod
+    def _resolve_field_value(
+        value: typing.Any, metadata: tuple[type["SimpleKeywordBox"]] | tuple[str, str]
+    ) -> typing.Any:
```
haneug: To be consistent with the rest of the codebase I would suggest to import Any from typing and just use Any? Or do you use typing.Any to stress that we are working with types here?

src/opi/simple_tasks/settings.py
```diff
@@ -0,0 +1,303 @@
...
+        """
+        Function to translate user input into OPI compatible types. This is done using the metadata of the type
+        annotation associated with the field. There are two cases:
+
+        Case 1
+        ------
+        Metadata has only one value. In this case the field is a simple keyword. Here the class associated with the input
```
haneug, line 160: Really like this docstring! Maybe some more explicit examples would be helpful. Since I do not know what goes in it is a bit difficult to grasp.

src/opi/simple_tasks/settings.py
```diff
@@ -0,0 +1,303 @@
...
+        -------
+        typing.Any
+            User input value translated to OPI compatible types.
+
+        """
+        match metadata:
+            case (validator,):
+                # Case 1: Simple keyword
+
+                if isinstance(validator, type) and issubclass(validator, Solvent):
```
haneug, line 203: These cases here are not directly obvious for me. First you check whether it is a `Solvent` which is not a Simple Keyword. If this is the case you just return the value (which is just the Solvent enum member I guess?)

src/opi/input/simple_keywords/base.py
```diff
@@ -1,28 +1,108 @@
...
+            elif simple_keyword.alias and any(a.lower() == norm for a in simple_keyword.alias):
+                return simple_keyword
+
+        # In the case that no matches are found, raise ValueError
+        raise ValueError(f"Keyword {s} not found in class {cls.__name__}")
+
+    @classmethod
+    def find_keyword(cls, inp: "SimpleKeyword | str") -> "SimpleKeyword":
+        """
+        Resolve a string or ``SimpleKeyword`` to a registered ``SimpleKeyword``.
```
haneug: The registry does no longer explicitly exist. Please update the docstring accordingly.

src/opi/simple_tasks/settings.py
```diff
@@ -0,0 +1,303 @@
...
+        ------
+        ValueError
+            If ``validator`` is ``SolvationModel`` but ``solvent`` is not set.
+        TypeError
+            If ``value`` is not a ``SolvationModelAndSolvent`` when required.
+        """
+        if validator == SolvationModel:
+            solvent_value = getattr(self, "solvent", None)
+            if solvent_value is None:
+                raise ValueError("solvent is required for SolvationModel")
```
haneug: A solvation model was requested without a solvent.

src/opi/simple_tasks/settings.py
```diff
@@ -0,0 +1,303 @@
...
+        The method does the following:
+        1. Retrieves type hints, including metadata and checks whether current field has corresponding type hint.
+        2. Extracts field specific metadata from type hint.
+        3. Resolves incoming value using metadata.
+
+        Parameters
+        ----------
+        value: Any
+            User input value.
+        info
```
haneug: info: ValidationInfo

src/opi/simple_tasks/settings.py
```diff
@@ -0,0 +1,303 @@
...
+        ----------
+        input_object: Input
+            `Input` object to be modified
+
+        Returns
+        -------
+        Input
+            Modified `Input` object.
+        """
+        hints = typing.get_type_hints(self.__class__, include_extras=True)
```
haneug: Here some more comments would be helpful I think. That is what I get here: You gather all the type hints. Then you get the value and the metadata for each field from the hints. Based on the metadata you decide whether it is a simple keyword or a bock to add. You loop over all the fields in the hints and add the corresponding simple keywords and blocks.

src/opi/simple_tasks/settings.py
```diff
@@ -0,0 +1,303 @@
...
+                case (validator,):
+                    if issubclass(validator, Solvent):
+                        continue
+
+                    new_keyword = self._get_simple_keyword(validator, value)
+                    if new_keyword:
+                        input_object.add_simple_keywords(new_keyword)
+
+                case (validator, key):
+                    block_type = Block.get_subclass_by_name(validator)
```
haneug: why do you have a dedicated function for the simple keywords but not for the blocks? I would suggest to add also a function for the blocks `_get_block`? The I would also appreciate some comments here what you are doing why. As far I understand you:
1. get the type of the block
2. initialize the block with the key: value and the block type
3. Check if the block exists and if not you add it. If it exists you merge them but how is not obvious to me.

src/opi/simple_tasks/settings.py
```diff
@@ -0,0 +1,303 @@
...
+                    block_instant = block_type(**{key: value})
+
+                    block_exists, *_ = input_object.has_blocks(block_type)  # type:ignore
```
haneug: why type ignore? Please add a comment if mypy was acting up here.

src/opi/simple_tasks/settings.py
```diff
@@ -0,0 +1,303 @@
...
+            if not isinstance(value, SolvationModelAndSolvent):
+                raise TypeError("Wrong type for solvent or solvation model")
+
+            new_keyword = value(solvent)
+        else:
+            new_keyword = validator.find_keyword(value)
+
+        return new_keyword
+
+    def __or__(self, other: "Settings") -> "Settings":
```
haneug: As a dunder method I would put this at the top since it describes the central behavior of `Settings`. After that I would put the `map_to_input` function as this is the central function of this class and after that put the helpers / validate methods.

src/opi/simple_tasks/method_settings.py
```diff
@@ -0,0 +1,399 @@
...
+            ``Input`` object to populate.
+
+        Returns
+        -------
+        Input
+            Modified ``Input`` object.
+        """
+        input_object = super().map_to_input(input_object)
+
+        if self.scf_stab:
```
haneug, line 259: Why is scf_stab added explicitly? Because the name does not fit?

src/opi/simple_tasks/method_settings.py
```diff
@@ -0,0 +1,399 @@
...
+    Exposes DLPNO-specific thresholds (``pno_thresh``, ``dlpno_t_cut_do``,
+    ``dlono_t_cut_pno``) and enables LED analysis via ``dlpno_led=True``.
+    SCF control knobs are also available.
+    """
+
+    model_config = ConfigDict(
+        arbitrary_types_allowed=True, validate_assignment=True, extra="forbid"
+    )
+    _name: str = "dlpnocc"
+    method: typing.Annotated[SimpleKeyword, Dft] | None = None
```
haneug: method should rather be `Wft` or only `DLPNO-CCSD(T)` and `DLPNO-CCSD` from Wft.

src/opi/simple_tasks/tasks/freq_task.py
```diff
@@ -0,0 +1,70 @@
...
+    def status(self) -> bool:
+        """``True`` if the job terminated normally (SCF convergence not re-checked)."""
+        return self.output.terminated_normally()
```
haneug, line 20: This just checks whether orca terminated normally and nothing related to the SCF. I also noted that checking if the SCF is converged is a status check you perform based on the method. So what status would actually have to do is perform checks based on the task and based on the method.

haneug (changes requested):
I looked through the settings class and I think it looks really nice. I would just add some more comments and flip the order of functions in the class. Also I noted that for status we will have method specific checks that need to be applied. E.g. Only for a method that has an SCF we need to check if the SCF is converged. We either have to check it for every method or we perform the health checks in a more general way e.g. first verify `has_scf` and the `scf_converged` such an automatic health check function in the output would also be of value outside of the simple tasks. I guess we discuss this in the JF.

src/opi/simple_tasks/settings.py
nakul680: It is merged using the __or__ operator that exists in Block class

src/opi/simple_tasks/settings.py
```diff
@@ -0,0 +1,330 @@
...
+                    # Two-element metadata: field maps to an attribute on a block
+                    block_type = Block.get_subclass_by_name(validator)
+                    block_instant = block_type(**{key: value})
+
+                    block_exists, *_ = input_object.has_blocks(block_type)  # type:ignore
+                    if not block_exists:
+                        input_object.add_blocks(block_instant)
+                    else:
+                        # Merge with any existing block of this type so no other attributes are lost
+                        existing_block = next(iter(input_object.get_blocks(block_type).values()))
```
haneug: Is the `next(iter(...))` construct really necessary here? since you have the type can''t you do `input_object.get_blocks(block_type)[block_type]`?

src/opi/simple_tasks/simple_task.py
```diff
@@ -0,0 +1,619 @@
...
+        filesystem.
+        """
+        if not self.calc_object:
+            raise ValueError("calc_object not set")
+
+        out = self.calc_object.get_output()
+        out.parse()
+        return out
+
+    @cached_property
```
haneug: These are still `@cached_property` but since they are inexpensive I think we decided to go with `@property`

src/opi/simple_tasks/simple_task.py
```diff
@@ -0,0 +1,619 @@
...
+            values from the shorthand parameters take precedence over fields
+            already present in ``method_settings``.
+
+        Raises
+        ------
+        ValueError
+            If neither ``method`` nor a ``method_settings`` object with a
+            method is provided.
+        """
+        task_settings_type = self._get_task_settings_type()
```
haneug: Please document here what you are doing any why with a few comments. Also a docstring for the init to explain the general plan here would be helpful.

src/opi/simple_tasks/simple_task.py
```diff
@@ -0,0 +1,619 @@
...
+        if isinstance(task_settings, dict):
+            self._task_settings = task_settings_type.model_validate(task_settings)
+        else:
+            self._task_settings = task_settings or task_settings_type()  # type: ignore[call-arg]
```
haneug: why the # type: ignore[call-arg] ?

haneug: Also how the or is resolved here is not directly obvious to me

src/opi/simple_tasks/simple_task.py
```diff
@@ -0,0 +1,619 @@
...
+
+        Raises
+        ------
+        AttributeError
+            If the current settings type does not have a ``method`` field.
+        """
+        if not hasattr(self._method_settings, "method"):
+            raise AttributeError("method is not defined in method_settings object")
+        if new_value is None:
+            self._method_settings.method = None
```
haneug, line 259: If it is None you just set it? Is there a specific reason? Please add a comment

src/opi/simple_tasks/simple_task.py
```diff
@@ -0,0 +1,619 @@
...
+            return
+        resolved_type = MethodSettings.resolve_method_settings_type(new_value)
+        if isinstance(self._method_settings, resolved_type):
+            self._method_settings.method = new_value  # type:ignore
```
haneug, line 264: why the type:ignore?

haneug (changes requested):
I added some smaller comments. Is somewhere described what validation is performed where. I find it a bit difficult to see that without any high-level overview.

src/opi/simple_tasks/settings.py
nakul680: True!

src/opi/simple_tasks/simple_task.py
nakul680: the ignore statement is there since in the original TaskSettings class definition , the task_keyword does not have a default value, initializing it here without the task_keyword defined runs into a mypy error. 

I will try making the task_keyword default in TaskSettings None, and if there are any problems because of it we can go back to the type:ignore statement.

haneug: Thanks for the explanation. In general I think it is good practice to explain these things so that when other people alter your code they are already warned about this.

PR #222, opened 2026-03-26
Minimal implementation of error_message

### LINKED ISSUE #215: Error message extraction
In case a calculation fails, it is nice to know why without spitting yourself through the output files.

It would be nice to have a method (`error_message()`?) which returns an error message, in the output module in case `output.terminated_normally()` returns `False`. Upon success this would just be an empty string, `None` would also do but not needed.

- Instead of a separate routine this could also be a field in an overloaded `terminated_normally(errmsg: str)` routine, but I think that is less clean.
- Using the grepper to spit myself through the output, but that is only a work-around in my opinion.

haneug:
## Closes Issues
Closes #215

## Description
- Implement a function `output.error_message` that goes through a list of known `ErrorPatterns` and checks if they are found in the output file. The pattern with the highest priority given by the position in the list will be returned. Tests for the errors can be found in `tests/failure_examples`. Initial errors include input errors, convergence errors for SCF, CC, geometry, and CP-SCF convergence, memory errors for SCF, MP2, and CC. 

---

### Added 

- Added `output.error_message()` which retrieves the most important error message from the ORCA output file (#222).
- Added `output.error_messages()` which retrieves all error messages from known strings from the ORCA output file (#222).
- Added `output.cc_converged` and `output.casscf_converged` to check cc and casscf convergence (#222).
- Added `z_maxiter` to method block for setting the number of CP-SCF iterations (#222).
 
### Fixed 
- Fixed a typo in the badge for the OPI paper (#222).


src/opi/input/blocks/block_method.py
```diff
@@ -127,6 +127,9 @@ class BlockMethod(Block):
     d3s8: float | None = None
     d3a2: float | None = None
 
+    # > Number of CPSCF iterations
+    z_maxiter: int | None = None
```
timmyte, line 131: I don't see this change listed in the PR description. Please cross-check.

src/opi/output/grepper/error_pattern.py
```diff
@@ -0,0 +1,152 @@
...
+import re
+from typing import Callable
+
+from opi.output.grepper.core import Grepper
+
+
+class ErrorPattern:
+    """
+    Represents an error pattern in the ORCA output file.
+    More complex error patterns derive from this class and override the extractor
```
timmyte:
```suggestion
    More complex error patterns derive from this class and override the `extractor`.
```

src/opi/output/grepper/error_pattern.py
```diff
@@ -0,0 +1,152 @@
...
+    critical: bool, default = False
+        When the error is critical we will stop searching for further errors after finding it.
+    extractor: Callable[[Grepper], str] | None, default = None
+        Optional function for extracting more details from the matched line.
+
+    """
+
+    def __init__(
+        self,
+        grep_string: str | None = None,
```
timmyte: Why not just go with empty strings as default? I don't see the value of using `None` here.
Applies to `grep_string` as well as `message`.

src/opi/output/grepper/error_pattern.py
```diff
@@ -0,0 +1,152 @@
...
+        message: str | None = None,
+        critical: bool = False,
+        extractor: Callable[[Grepper], str] | None = None,
+    ) -> None:
+        if grep_string is not None:
```
timmyte: I would check as follows: `if not grep_string` (also applies to `if message ...`)
E.g. why would one want grep for an empty string?

src/opi/output/grepper/error_pattern.py
```diff
@@ -0,0 +1,152 @@
...
+            self.grep_string = grep_string
+        if message is not None:
+            self.message = message
+        if critical is not None:
```
timmyte: Above it says `critical` is always boolean. So I don't get why you check for `None` here.

src/opi/output/grepper/error_pattern.py
```diff
@@ -0,0 +1,152 @@
...
+            self.critical = critical
+        self.extractor = extractor
+
+    def match(self, grepper: Grepper) -> str | None:
```
timmyte: 1) Again. I would stick with an empty string rather than `None` as return value.
Also applies to the `extract()` method.
2) Please add docstring. Especially in the base class. And also some comments please.

src/opi/output/grepper/error_pattern.py
```diff
@@ -0,0 +1,152 @@
...
+        hit = grepper.search(self.grep_string, case_sensitive=True)
+        if not hit:
+            return None
+        if self.extractor:
+            return self.extractor(grepper)
+        return self.extract(grepper) or self.message
+
+    def extract(self, grepper: Grepper) -> str | None:
```
timmyte: This method should have a docstring. Especially in the base class.

src/opi/output/grepper/error_pattern.py
```diff
@@ -0,0 +1,152 @@
+# patterns.py
```
timmyte: This is not done consistently adding the filename in the first line as comment.
Either do it for all files or none (which I would prefer), but please don't establish such new conventions in some (random) files only.

src/opi/output/grepper/error_pattern.py
```diff
@@ -0,0 +1,152 @@
...
+        if grep_string is not None:
+            self.grep_string = grep_string
+        if message is not None:
+            self.message = message
+        if critical is not None:
+            self.critical = critical
+        self.extractor = extractor
+
+    def match(self, grepper: Grepper) -> str | None:
+        hit = grepper.search(self.grep_string, case_sensitive=True)
```
timmyte: I actually don't see the point why we would first search for `self.grep_string` and later use the `extractor` which looks for the same string.
Can we just process the return value of the `extractor` to determine if the error pattern applies here?

src/opi/output/core.py
```diff
@@ -672,6 +675,13 @@ def terminated_normally(self) -> bool:
         except FileNotFoundError:
             return False
 
+    def error_message(self) -> str | None:
+        outfile = self.get_outfile()
+        try:
+            return get_error_message(outfile)
+        except FileNotFoundError:
+            return "Output File Not Found"
```
timmyte: I would add the expected path ;)

src/opi/output/core.py
```diff
@@ -689,6 +699,40 @@ def scf_converged(self) -> bool:
...
+        Returns
+        -------
+        bool
+            True if string is found in ".out" file else False
+        """
+        outfile = self.get_outfile()
+        try:
+            return has_casscf_converged(outfile)
+        except FileNotFoundError:
+            return False
```
timmyte: These lines seem to develop into a repeating pattern. We should write a driver for *has_xxx*-function. Something like:
```
# > Function name just exemplary
def _has(self, has_func: Callable[[Path], bool]) -> bool:
        outfile = self.get_outfile()
        try:
            return has_func(outfile)
        except FileNotFoundError:
            return False
```
`casscf_converged()` would then look as follows:
```
def casscf_converged(self):
     return self._has(has_cc_converged)
```

tests/failure_examples/test_convergence_failure.py
```diff
@@ -0,0 +1,86 @@
+#!/usr/bin/env python3
+import pytest
+
+from opi.core import Calculator
+from opi.input.blocks import BlockGeom, BlockMdci, BlockMethod, BlockScf
+from opi.input.simple_keywords import AuxBasisSet, Task, Wft
+from opi.input.structures import Structure
+
+"""
+Contains ORCA examples of convergence failures to test OPI´s error handling capabilities.
```
timmyte: I would speak of `error extraction capabilities`

tests/failure_examples/test_convergence_failure.py
```diff
@@ -0,0 +1,86 @@
...
+The functions error_message() will search in the ORCA output file for a respective error string and will compose the
```
timmyte:
```suggestion
The function `error_message()` will search in the ORCA output file for a series of common error messages by pattern and compose the
```
To me this reads as if it should rather go into the docstring of `error_message()` directly (except for the last part about *asserting*)

tests/failure_examples/test_convergence_failure.py
```diff
@@ -0,0 +1,86 @@
...
+
+@pytest.fixture
+def calc(tmp_path):
+    """Create a calculator object with a water structure and return it."""
+    calc = Calculator(basename="job", working_dir=tmp_path)
+    calc.structure = Structure.from_smiles("O")
+    return calc
+
+
+@pytest.mark.orca
```
timmyte, line 29: Are you sure that this the only test that should be marked with `orca`?

tests/failure_examples/test_input_failure.py
```diff
@@ -0,0 +1,125 @@
+#!/usr/bin/env python3
+import pytest
+
+from opi.core import Calculator
+from opi.input.blocks import BlockScf
+from opi.input.structures import Structure
+
+"""
+Contains ORCA examples of input failures to test OPI´s error handling capabilities.
```
timmyte: Exactly the same module documentation as above. Please be at least a bit more creative.

tests/failure_examples/test_convergence_failure.py
```diff
@@ -0,0 +1,86 @@
+#!/usr/bin/env python3
+import pytest
+
+from opi.core import Calculator
+from opi.input.blocks import BlockGeom, BlockMdci, BlockMethod, BlockScf
+from opi.input.simple_keywords import AuxBasisSet, Task, Wft
+from opi.input.structures import Structure
+
+"""
```
timmyte: Module documentation should come right after the shebang and before any Python code (including imports).

tests/failure_examples/test_memory_failure.py
```diff
@@ -0,0 +1,67 @@
+#!/usr/bin/env python3
+import pytest
+
+from opi.core import Calculator
+from opi.input.blocks import BlockMdci
+from opi.input.simple_keywords import BasisSet, Scf, Wft
+from opi.input.structures import Structure
+
+"""
+Contains ORCA examples of memory failures to test OPI´s error handling capabilities.
```
timmyte: Must be a glitch in the matrix...

tests/failure_examples/test_convergence_failure.py
```diff
@@ -0,0 +1,86 @@
...
+def test_scf_conv_fail(calc):
+    """Test error_message for SCF failure"""
+    calc.input.add_blocks(BlockScf(maxiter=1))
+    # > write the input and run the calculation
+    calc.write_and_run()
+
+    # > get the output and check some results
+    output = calc.get_output()
+    assert not output.terminated_normally()
+    assert output.error_message() == "SCF did not converge"
```
timmyte: Instead of hard-coding the expected return string here, I would rather compare against the string stored in the corresponding `ErrorPattern` itself.

tests/failure_examples/test_memory_failure.py
```diff
@@ -0,0 +1,67 @@
...
+    calc.input.add_simple_keywords(BasisSet.DEF2_QZVPPD)
+    # > write the input and run the calculation
+    calc.write_and_run()
+
+    # > get the output and check some results
+    output = calc.get_output()
+    print(output.error_message())
+    assert not output.terminated_normally()
+    assert (
+        "Not enough memory available for SCF. Available:           1.0 MB" in output.error_message()
```
timmyte: I would be careful with using a fixed number of multiple consecutive blanks in a string literal if it's not meant only for printing. This where regex should be used.

README.md
```diff
@@ -4,7 +4,7 @@
 ![Static Badge](https://img.shields.io/badge/contributing-CLA-red)
 ![Static Badge](https://img.shields.io/badge/DOI-10.5281%2Fzenodo.15688425-blue)
 ![Static Badge](https://img.shields.io/badge/release-2.0.0-%2300AEC3)
-[![JCTC Paper](https://img.shields.io/badge/JCTC-10.1021%2Faces.jctc.5c02141-blue?logo=doi)](https://doi.org/10.1021/acs.jctc.5c02141)
+[![JCTC Paper](https://img.shields.io/badge/JCTC-10.1021%2Facs.jctc.5c02141-blue?logo=doi)](https://doi.org/10.1021/acs.jctc.5c02141)
```
timmyte, line 7: Is this change listed in the PR description?

src/opi/output/grepper/patterns.py
```diff
@@ -0,0 +1,80 @@
...
+
+# > Has strings
+HAS_GEOMETRY_OPT = "Geometry Optimization Run"
+HAS_SCF = "SCF SETTINGS"
+HAS_ABORTING = "aborting"
+
+# > Error patterns in order of priority.
+# > Critical errors will stop scanning when matched.
+# > Non-critical errors will just be added and reported.
+ERROR_PATTERNS: list[ErrorPattern] = [
```
timmyte, line 89: Please add kind of *line* in the form of a comment, at which point in the list we go from critical to non-critical.

src/opi/output/grepper/recipes.py
```diff
@@ -1,6 +1,36 @@
...
+    for pattern in ERROR_PATTERNS:
+        msg = pattern.match(grepper)
+        if msg:
+            hits.append(msg)
+            if pattern.critical:
+                break
+    return hits if hits else None
+
+
+def get_error_message(file_name: Path) -> str | None:
```
timmyte: I would make the function name more specific: e.g. `get_first_error_message()`

src/opi/output/grepper/patterns.py
```diff
@@ -0,0 +1,80 @@
+from opi.output.grepper.error_pattern import (
```
timmyte, line 6: Please add module level docstring, describing what this file holds.

timmyte (changes requested):
This is nice improvement in the right direction. This will allow users to all add fallbacks to there OPI scripts.

But please cross-check the release notes and description again and removed sections, that are not served in this PR.

Also remove the three dots in the "Closes Issues" section.

src/opi/output/grepper/recipes.py
haneug: I would not like to do that since the name becomes quite long and the intention of `error_message` is not to get the first one but the most important / most specific one.

haneug:
@timmyte Sorry for keeping this laying around for so long. I finally could clean this up and I think it is ready for a new round of reviews

src/opi/output/grepper/error_pattern.py
```diff
@@ -0,0 +1,157 @@
...
+        The string that is searched in the output file.
+    message: str
+        A human-readable error message of the given error pattern.
+    critical: bool
+        When the error is critical we will stop searching for further errors after finding it.
+        Critical errors are errors after which ORCA will abort.
+
+    """
+
+    grep_string: str = ""
```
timmyte, line 24: I wouldn't define these variables as class variables, and then in the constructor override these variables only if the corresponding arguments have non-default values.
What actually happens here, is that if `self.grep_string = grep_string` is triggered, second variable `grep_string` is defined as instance variable, which has precedence over class variables.

The constructor should looks as follows:
```
 def __init__(
        self,
        grep_string: str = "",
        message: str = "",
        critical: bool = False,
    ) -> None:
        self.grep_string = grep_string if grep_string else ""
        if message:
        self.message = message if message else ""
        self.critical = critical
```

Also notice for the boolean `critical`:
```
if critical:
    self.critical = critical
```
This construct assumes the default is `False`. But I wouldn't make the assumption here. What if the default is changed in the future to `True`, then `self.critical` could never be turned off in the constructor.

timmyte (approved):
Solid work!
This will be a nice addition to the API.

Please fix my one remark before merging. Thanks!!

PR #224, opened 2026-03-31
Orca MM Runner

### LINKED ISSUE #185: Running Orca MM
It would be nice to be able to run the `orca_mm` program from `opi`.

`orca_mm` has many sub-commands, the main one I am currently using is `convff` but a solution should be to run any of the `orca_mm` sub-commands. I can see two methods of implementing this, the first is having all the commands behind a single method, the other is to split up each sub-command into an individual method.

**Single Method**
```python
OrcaMMCommand = Literal["convff", "splitff", "mergeff", "repeatff", "splitpdb", "mergepdb", "makeff", "getHDist"]
OrcaFFType = Literal["amber", "charmm", "openmm"]

class Runner:
    ...
    def run_orca_mm_convff(
        self,
        command: OrcaMMCommand,
        /,
        *,
        fftype:  OrcaFFType | None = None,
        ffinput: Path | list[Path] | None = None,
        // etc.
        silent: bool = True,
        timeout: int = -1,
    ): ...
```

**Multiple Methods**
```python
class Runner:
    ...
    def run_orca_mm_convff(self, fftype:  OrcaFFType, ffinput: Path | list[Path]): ...
    def run_orca_mm_splitff(self, ffinput: Path, atoms: list[int]): ...
    def run_orca_mm_splitff(self, ffinputs: list[Path]): ...
```

Another option is to subclass `Runner` into `OrcaMmRunner` so that having multiple methods for each sub-command does not clutter the main `Runner` class.

MattBurn:
## Closes Issues
Closes #185

## Description
Adds `OrcaMmRunner` in `opi.execution.orca_mm` for executing supported `orca_mm`  commands through `opi`.

The runner provides typed helpers for:

- `orca_mm -convff`
- `orca_mm -splitff`
- `orca_mm -mergeff`
- `orca_mm -repeatff`
- `orca_mm -splitpdb`
- `orca_mm -mergepdb`
- `orca_mm -makeff`
- `orca_mm -getHDist`

Each command-specific method validates its input files and arguments, assembles the appropriate command line, executes `orca_mm`, and verifies that the expected output files were generated.

The lower-level `run_orca_mm()` method builds on the updated `BaseRunner.run()` API introduced in #238. It supports `stdout` and `stderr` stream targets, returns a `RunResult`, and raises `OrcaMmError` by default when `orca_mm` exits with a nonzero return code. STDERR is captured when required so that it can be included in the error message while still being forwarded to any caller-provided stream targets.

---

### Added

- Added `OrcaMmRunner` with typed helpers for supported `orca_mm` commands. (#224)
- Added `OrcaBinary.ORCA_MM`. (#224)
- Added `OpiError`, `OpiExecutionError`, and `OrcaMmError`. (#224)
- Added input-file validation and expected-output verification for `orca_mm` command helpers. (#224)


src/opi/execution/base.py
timmyte:
```suggestion
    Base class that facilitates the execution of ORCA binaries, especially the main ORCA binary.
```

src/opi/execution/orca_mm.py
```diff
@@ -0,0 +1,539 @@
...
+    "PBEOptH",
+    "XTB",
+    "XTBOpt",
+    "XTBOptH",
+    "XTBOptPBE",
+    "noChargeCalc",
+]
+
+
+class OrcaMmException(Exception):
```
timmyte: @haneug 
If we start implementing custom exceptions, we should have base class `OpiError` from which all custom exceptions are derived.

src/opi/execution/mm.py
```diff
@@ -0,0 +1,539 @@
...
+    Return a sibling path with a suffix-preserving name replacement.
+
+    Parameters
+    ----------
+    path : Path
+        Source file path.
+    new : str
+        String inserted before `suffix` in the filename.
+    suffix : str
+        Suffix that is replaced in the original filename.
```
timmyte: I would rather just speak of: `Suffix of file path`

src/opi/execution/orca_mm.py
```diff
@@ -0,0 +1,539 @@
...
+                OrcaBinary.ORCA_MM,
+                [f"-{command}"] + list(arguments),
+                stderr=path,
+                silent=silent,
+                timeout=timeout,
+            )
+            if raise_on_error and path.exists() and (error := path.read_text()):
+                raise OrcaMmException(command, arguments, error)
+
+    def run_convff(
```
timmyte, line 153: Please document on class level, that there is also a method per command of `orca_mm` as this is this a repeating design pattern.

src/opi/execution/base.py
timmyte: Not exactly. The class itself already allows running any ORCA binary, but for smoother a API it is helpful to subclass it.
Please reflect that in the documentation. Thanks

src/opi/execution/mm.py
```diff
@@ -0,0 +1,539 @@
...
+        """
+        fmt_arguments = " ".join(arguments)
+        message = (
+            f"Failed running command: 'orca_mm -{command} {fmt_arguments}'.\n"
+            f"orca_mm failed with the following error:\n{error.strip()}"
+        )
+        super().__init__(message)
+
+
+def _replace_suffix(path: Path, new: str, suffix: str) -> Path:
```
timmyte: The function name is misleading as the suffix is preserved.
How about: `_prepend_suffix` or `_add_infix`.

src/opi/execution/mm.py
```diff
@@ -0,0 +1,539 @@
...
+    """
+    Return a sibling path with a suffix-preserving name replacement.
```
timmyte: This is very technical. How about: `Return a sibling path with the name replaced while preserving a suffix.`

Further, try to distance this function from `Path.with_stem()` in the docstring, as `Path` defines suffix a bit more strictly.

src/opi/execution/mm.py
```diff
@@ -0,0 +1,539 @@
...
+        """
+        try:
+            yield
+        except Exception:
+            raise
+
+        missing_outputs = [path for path in expected_outputs if not path.exists()]
+        if len(missing_outputs) == 0:
+            return
+        elif len(missing_outputs) == 1:
```
timmyte: I wouldn't treat the single-file-missing-case different from multiple files missing. The underlying issue is still the same. This makes grepping for such strings easier in batch execution.

src/opi/execution/orca_mm.py
```diff
@@ -0,0 +1,539 @@
...
+            Capture and discard STDOUT and STDERR.
+        timeout : int, default: -1
+            Optional timeout in seconds to wait for process to complete.
+
+        Raises
+        ------
+        OrcaMmException
+            If `raise_on_error` is set and `orca_mm` reports an error.
+        """
+        with tempfile.TemporaryDirectory() as tmp_dir:
```
timmyte, line 141: I don't get what we need the temporary dir for? Just to dump the STDERR somewhere?
Why capture via `subprocess.PIPE`? This could be easily added to `self.run()`.

src/opi/execution/mm.py
```diff
@@ -0,0 +1,539 @@
...
+        self,
+        command: OrcaMmCommand,
+        arguments: Sequence[str],
+        *,
+        raise_on_error: bool = True,
+        silent: bool = True,
+        timeout: int = -1,
+    ):
+        """
+        Execute `orca_mm` with the provided command and arguments.
```
timmyte:
```suggestion
        Execute `orca_mm` with the provided subcommand and arguments.
```

src/opi/execution/mm.py
```diff
@@ -0,0 +1,539 @@
...
+        ffinput: ForcefieldType,
+        ff_files: Iterable[Path],
+        *,
+        force: bool = False,
+        raise_on_error: bool = True,
+        silent: bool = True,
+        timeout: int = -1,
+    ) -> Path:
+        """
+        Executes the orca_mm binary with the -convff flag and passes in the forcefield type
```
timmyte:
```suggestion
        Executes the `orca_mm` binary with the `-convff` flag and passes in the forcefield type
```
Please wrap `orca_mm` consistently in backticks.

src/opi/execution/orca_mm.py
```diff
@@ -0,0 +1,539 @@
...
+        ffinput : ForcefieldType
+            Input forcefield format (`amber`, `charmm`, or `openmm`).
+        ff_files : Iterable[Path]
+            Input forcefield file(s) to convert.
+        force : bool, default: False
+            Overwrite existing output file if present.
+        raise_on_error : bool, default: True
+            Raise `OrcaMmException` if `orca_mm` reports an error.
+        silent : bool, default: True
+            Capture and discard STDOUT and STDERR.
```
timmyte, line 178: 1) OK but what if `silent` and `raise_on_error` are both `True` and STDOUT and STDERR are discarded. How would detect an error and where do we fetch the error message from?
2) What if `silent=False` and `raise_on_error=True`. Isn't STDERR then not also discarded?

Please cover these two points in all docstrings containing the two parameters.

src/opi/execution/orca_mm.py
```diff
@@ -0,0 +1,539 @@
...
+        -------
+        Path
+            Path to generated ORCA forcefield file (`*.ORCAFF.prms`).
+        """
+        ff_files = list(ff_files)
+
+        if len(ff_files) == 0:
+            raise ValueError("Must supply at least 1 forcefield file.")
+
+        expected_output = ff_files[0].with_suffix(self._orca_ff_suffix)
```
timmyte, line 197: Please add a comment here, what this section does.

src/opi/execution/orca_mm.py
```diff
@@ -0,0 +1,539 @@
...
+        if expected_output.is_file() and not force:
+            return expected_output
+
+        expected_output.unlink(missing_ok=True)
```
timmyte, line 204: Please add a comment here, why delete the file here?

src/opi/execution/orca_mm.py
```diff
@@ -0,0 +1,539 @@
...
+
+        arguments = [f"-{ffinput}"] + [str(f) for f in ff_files]
```
timmyte, line 206: Please add a comment here, what this section does and why.

src/opi/execution/orca_mm.py
```diff
@@ -0,0 +1,539 @@
...
+        """
+        sorted_atoms = list(sorted(atoms))
+
+        if len(sorted_atoms) == 0:
+            raise ValueError("Must supply at least 1 atom.")
+
+        if any(atom < 1 for atom in sorted_atoms):
+            raise ValueError("All atoms must be positive integers.")
+
+        expected_outputs = [
```
timmyte, line 251: Please add a comment here, what this section does and why.

src/opi/execution/orca_mm.py
```diff
@@ -0,0 +1,539 @@
...
+            _replace_suffix(orcaff_file, f"_split{split + 1}", self._orca_ff_suffix)
+            for split in range(len(sorted_atoms) + 1)
+        ]
+
+        arguments = [f"{orcaff_file}"] + [str(atom) for atom in sorted_atoms]
```
timmyte, line 256: Please add a comment here, what this section does and why.

src/opi/execution/mm.py
```diff
@@ -0,0 +1,539 @@
...
+
+        Returns
+        -------
+        Path
+            Path to merged ORCA forcefield file.
+        """
+        if len(orcaff_files) < 2:
+            raise ValueError("Must provide at least 2 orca ff files to merge")
+
+        expected_output = _replace_suffix(orcaff_files[0], "_merged", self._orca_ff_suffix)
```
timmyte: Please add a comment here, what this section does and why.

src/opi/execution/orca_mm.py
```diff
@@ -0,0 +1,539 @@
...
+        expected_output = ff_files[0].with_suffix(self._orca_ff_suffix)
+        if expected_output.is_file() and not force:
+            return expected_output
+
+        expected_output.unlink(missing_ok=True)
+
+        arguments = [f"-{ffinput}"] + [str(f) for f in ff_files]
+        with self._expect_output_files(expected_output):
+            self.run_orca_mm(
+                "convff", arguments, raise_on_error=raise_on_error, silent=silent, timeout=timeout
```
timmyte, line 209: Docstring of the function says `-convff` with a leading dash. Please be consistent. Either fix the documentation or the call args here.

Please do this for all function in this module!

src/opi/execution/orca_mm.py
```diff
@@ -0,0 +1,539 @@
...
+                silent=silent,
+                timeout=timeout,
+            )
+            if raise_on_error and path.exists() and (error := path.read_text()):
+                raise OrcaMmException(command, arguments, error)
+
+    def run_convff(
+        self,
+        ffinput: ForcefieldType,
+        ff_files: Iterable[Path],
```
timmyte, line 156: I would also add checks that each input file exists. Please do this for all methods.
Could be done with simple method:
```
def _assert_files_exist(files: Path, /) -> None:
for file in files:
    if not file.is_file():
        raise FileNotFoundError(...)
```

src/opi/execution/orca_mm.py
```diff
@@ -0,0 +1,539 @@
...
+        Path
+            Path to repeated ORCA forcefield file.
+        """
+        if repeat < 1:
+            raise ValueError("'repeat' must be a positive integer")
+
+        expected_output = _replace_suffix(orcaff_file, f"_repeat{repeat}", self._orca_ff_suffix)
+
+        arguments = [f"{orcaff_file}", str(repeat)]
+        with self._expect_output_files(expected_output):
```
timmyte, line 339: I really like the idea with the context manager. Very smooth.
Just one minor improvement suggestion: The context manager followed by the call to `run_orca_mm()` are a repeating pattern.
I would wrap this in another method:
```
def run_orca_mm_and_expect(expected_files: Path | Sequence[Path], *args, **kwargs)
# *args and **kwargs go into self.run_orca_mm()
```

src/opi/execution/orca_mm.py
```diff
@@ -0,0 +1,539 @@
...
+        """
+        sorted_atoms = list(sorted(atoms))
+
+        if len(sorted_atoms) == 0:
+            raise ValueError("Must supply at least 1 atom.")
+
+        if any(atom < 1 for atom in sorted_atoms):
+            raise ValueError("All atoms must be positive integers.")
+
+        expected_outputs = [
```
timmyte, line 387: Please add a comment here, what this section does and why.

src/opi/execution/orca_mm.py
```diff
@@ -0,0 +1,539 @@
...
+            Optional timeout in seconds to wait for process to complete.
+
+        Returns
+        -------
+        Path
+            Path to generated ORCA forcefield file (`*.ORCAFF.prms`).
+        """
+        expected_output = structure_file.with_suffix(self._orca_ff_suffix)
+
+        arguments = [str(structure_file)]
```
timmyte, line 483: Please add some more comments to this routine, to point out the individual steps and preparations done here, like,
1) generating files
2) assembling call
3) performing call

tests/unit/test_execution_orca_mm.py
```diff
@@ -0,0 +1,50 @@
+from pathlib import Path
```
timmyte, line 1: Please add a module level description what is tested in this module and also list the individual tests.

tests/unit/test_execution_orca_mm.py
```diff
@@ -0,0 +1,50 @@
...
+
+import pytest
+
+from opi.execution.mm import OrcaMmException, OrcaMmRunner
+
+
+def _set_fake_orca_path(self, orca_path: Path | None = None) -> None:
```
timmyte, line 8: Please add docstrings.
Mention why these *fake* are necessary as this is not trivial.

tests/unit/test_execution_orca_mm.py
```diff
@@ -0,0 +1,50 @@
...
+    self._orca_lib_folder = Path("/tmp")
+
+
+def _set_fake_open_mpi_path(self, mpi_path: Path | None = None) -> None:
+    self._open_mpi_path = None
+
+
+@pytest.mark.unit
+def test_run_orca_mm_handles_deleted_empty_stderr(monkeypatch: pytest.MonkeyPatch):
+    monkeypatch.setattr(OrcaMmRunner, "set_orca_path", _set_fake_orca_path)
```
timmyte, line 19: Didn't know Pytest has `MonkeyPatch`. Very sophisticated. Thanks!!

tests/unit/test_execution_orca_mm.py
```diff
@@ -0,0 +1,50 @@
...
+    monkeypatch.setattr(OrcaMmRunner, "set_open_mpi_path", _set_fake_open_mpi_path)
+
+    runner = OrcaMmRunner()
+
+    def fake_run(binary, args, /, *, stderr=None, silent=True, timeout=-1):
```
timmyte, line 24: Again, please add docstring. Why *fake* execution. Please describe your design. Could also do this on module level, as the pattern is the same.

timmyte (dismissed):
@MattBurn 
Thanks a lot for the PR. Thanks a very nice addition and improvement for OPI. Also thanks to directly include proper tests.

I would appreciate if you could overall add more comments and documentation to the code. Most patterns are very repetitive, so they would only a single documentation somewhere prominently on module level.

src/opi/execution/orca_mm.py
```diff
@@ -0,0 +1,539 @@
+"""
```
timmyte, line 1: Also, please give the module the same name as the binary it executes/wraps. Thanks.

src/opi/execution/orca_mm.py
MattBurn: What path would you like to take? I can either change all OrcaMmExceptions into built in exception types or add an OpiError/OpiException if the latter do you prefer exception or error?

src/opi/execution/orca_mm.py
MattBurn: does this also apply to the sore orca runner or just to `orca mm`? What would you like the module to be called? `orca_mm.py` , `mm_runner.py`, `orca_mm_runner.py`?

src/opi/execution/orca_mm.py
MattBurn: @timmyte @haneug 

I have added the requested comments but before I commit I'd like to get your opinion on the expected behaviour of this function. `orcamm -convff` generates a file with the same stem as the first forcefield file provided and the suffix replaced with '.ORCAFF.prms'. This means if I run the function twice with the same arguments then the orcaff file will already exist on the second invocation.

The reason I delete an existing file first is so that we can check the `expected_output` has actually been generated. If the `expected_output` already exists and we aren't running with `raise_on_error` then we could end up in the situation where `orca_mm` fails but the output file is still there so the check passes.

Ideally I would like a safer API  where any error running an orca binary raises an exception by default.

This would change the behaviour of the base orca runner which is why I restricted this behaviour to the `OrcaMmRunner` for now.

This also touches on another comment you have made regarding the incompatibility between the `raise_on_error` and `silent` arguments. Ideally I would like to refactor the `BaseRunner.run` API to remove `silent`, always capture `stdout` and `stderr` internally and allow a user to pass in a Path or TextIO which we can also optionally pipe the output to.

I believe this logic would simplify how `opi` is currently handling interacting with the subprocess. Let me know your thoughts.

src/opi/execution/orca_mm.py
haneug: I generally like the idea of adding `OpiError`. We could have a small implementation in `opi/exceptions.py` and derive `OpiExecutionError` from it and `OrcaMmError` would then be a subclass of that I guess. Would that be okay with you @timmyte ?

src/opi/execution/orca_mm.py
timmyte: I would go with `orca_mm.py`.
```
from opi.execution.orca_mm import ...
```
looks intuitive to me.

src/opi/execution/orca_mm.py
MattBurn: Do we want `OpiError` and `OpiExecutionError` to exist in `opi/exceptions.py` or put `OpiExecutionError` into `opi/execution/base.py` (or even it's own `opi/execution/exceptions.py`)?

src/opi/execution/orca_mm.py
MattBurn: My apologies, I didn't grasp the idea behind `capture=True` when implementing this and thought my only control was by passing in a `Path`. I have detailed my thoughts on the implementation of `BaseRunner.run` in issue #235

src/opi/execution/orca_mm.py
MattBurn: I have detailed my thoughts on `BaseRunner.run` in issue #235

src/opi/execution/orca_mm.py
MattBurn: I decided to create an issue related to this topic (#235)

src/opi/execution/orca_mm.py
MattBurn: The `-convff` in the docstring is related to the `orca_mm` binary earlier in the same sentence of the docstring. Apologies if this was unclear. Do you think that the `command` in `run_orca_mm` should already have the `-` prefixed so that there is not a mismatch between the python functions and what the binary is running or just make it clearer in the docstrings?

src/opi/execution/orca_mm.py
timmyte: Sorry for my late response.

> The reason I delete an existing file first is so that we can check the `expected_output` has actually been generated. If the `expected_output` already exists and we aren't running with `raise_on_error` then we could end up in the situation where `orca_mm` fails but the output file is still there so the check passes.

Good catch. Agreed!
Also having the `raise_on_error` parameter that is by default `True` is nice a addition. It would be appreciated, if you could add this to the other parts in the `[Base]Runner` infrastructure as well.

> This also touches on another comment you have made regarding the incompatibility between the `raise_on_error` and silent arguments. Ideally I would like to refactor the `BaseRunner.run` API to remove silent, always capture `stdout` and `stderr` internally and allow a user to pass in a Path or TextIO which we can also optionally pipe the output to.

Fine by me. We currently suppress/redirect all output for every use of `run()` anyway. But then we would also have to get rid of `capture` as well. `capture=True` would be implied and if `stdout` and/or `stderr` are passed with a file path, then the streams are independently redirected.
However, I wouldn't allow the user to pass their own string buffer. As the streams are captured by default anyway and returned as `subprocess.CompletedProcess[str]` already.

Does this align with your vision for the `Runner`?

MattBurn: Yes, kind of, but I believe there are some use cases that can be improved by allowing passing in a more flexible object for stdout and stderr.

Have a look at the linked issue as I think this could do with its own PR.

src/opi/execution/orca_mm.py
timmyte: I would follow @haneug's suggestion to have them in a central place in `opi/exceptions.py`.

src/opi/execution/orca_mm.py
MattBurn: This has now been removed in favor of capturing `stderr` output using `subprocess.PIPE`

src/opi/execution/orca_mm.py
MattBurn: This should be resolved by using the new `BaseRunner.run` `stderr` target. The `silent` flag has been removed and we use the process exit code to determine whether an error occurred whilst capturing `stderr` to use in the error message.

src/opi/execution/orca_mm.py
MattBurn: Added some clarifying comments around the `unlink` statement. We could still have `raise_on_error=False` which would prevent us from checking whether the new output has been generated (without using timestamps).

An alternative behavior would be to raise an error if the expected output already exists and `force` is `False` instead of just returning the path that already exists.

src/opi/execution/orca_mm.py
MattBurn: I have added a clarifying comment.

src/opi/execution/orca_mm.py
MattBurn: I have added a clarifying comment.

src/opi/execution/mm.py
MattBurn: I have added a clarifying comment.

src/opi/execution/orca_mm.py
MattBurn: I have wrote in the docstring that the function executes the `orca_mm` binary with the `-convff` flag. This is consistent with the other docstrings in the file.

src/opi/execution/orca_mm.py
MattBurn: Added as a `staticmethod` on the class, let me know if you think this should be a function outside of the class.

src/opi/execution/orca_mm.py
MattBurn: Implemented the helper method `_run_orca_mm_and_expect(...)`

src/opi/execution/orca_mm.py
MattBurn: Added the clarifying comments

tests/unit/test_execution_orca_mm.py
MattBurn: Replaced these methods with a fixture, added docstring to the new fixture and a test module docstring

tests/unit/test_execution_orca_mm.py
MattBurn: Added docstrings to the new fixture

tests/unit/test_execution_orca_mm.py
MattBurn: updated the test level monkeypatching with a reusable `orca_mm_runner_factory` fixture which allows for setting up a mocked runner and returns the `Mock` object so that we can interrogate what was passed into the `BaseRunner.run` method.

tests/unit/test_execution_orca_mm.py
MattBurn: done

src/opi/execution/orca_mm.py
MattBurn: done

src/opi/execution/orca_mm.py
MattBurn: done

src/opi/execution/orca_mm.py
MattBurn: done

MattBurn:
With the newly merged #235, I rebased this branch to allow the use of the new `BaseRunner.run` method. This helped to make the implementation of `OrcaMmRunner` simpler and removing the ambiguous behavior between `silent=True` and `raise_on_error=True`.

I have resolved most comments from the previous review, if you are not happy with any of the resolutions feel free to let me know. I have left some comments unresolved:
- [Removing existing files](https://github.com/faccts/opi/pull/224#discussion_r3155618905)
  - I have added the clarifying comments on why we need to remove the existing file so that we can verify the `orca_mm` command produced our expected output. I am still unsure of whether returning the expected output path if the file exists is the safest option and whether you have an opinion on the matter.
- [Adding a file check](https://github.com/faccts/opi/pull/224#discussion_r3064837836)
  - I added this as a `staticmethod` but could be moved to a module level helper function (stylistic choice I am fine with either)
- [Docstring leading dash](https://github.com/faccts/opi/pull/224#discussion_r3064824018)
  - I have been consistent throughout the docstrings regarding using the leading dash when referencing which `orca_mm` command the function will run. I do this because I am always referencing against the `orca_mm` binary rather than the `OrcaMmRunner.run_orca_mm` method which does not require the leading dash. All docstrings are consistent against referencing the underlying `orca_mm` command but I can see how that could lead to confusion regarding the use of the `OrcaMmRunner.run_orca_mm` function. I'd be happy updating this if you feel like it needs it.

Would I be able to request a new review to make sure I have properly resolved the comments from the previous review and to check through the integration with the new `BaseRunner.run` method. So far the test suite for `OrcaMmRunner` is quite light so if you feel like more testing is required then just let me know.

tests/unit/test_execution_orca_mm.py
```diff
@@ -0,0 +1,127 @@
...
+
+@pytest.mark.unit
+def test_run_orca_mm_raises_on_nonzero_returncode(
+    orca_mm_runner_factory: RunnerFactory,
+) -> None:
+    """Test that unsuccessful binary execution raises `OrcaMmError`."""
+    runner, _ = orca_mm_runner_factory(returncode=1, stderr="orca_mm failed")
+
+    with pytest.raises(OrcaMmError, match="orca_mm failed"):
+        runner.run_orca_mm("convff", ["-amber", "test.prm"])
```
nakul680, line 96: Im wondering if this test should include a more specific field for `stderr`. Since the error is reported as:

`f"orca_mm failed with the following error:\n{error.strip()}"`

the test would pass regardless of what string value is set for stderr. A more specific error text would make for a better test

src/opi/execution/orca_mm.py
```diff
@@ -0,0 +1,843 @@
...
+        ----------
+        *files : Path
+            Input file paths to validate.
+
+        Raises
+        ------
+        FileNotFoundError
+            If one or more paths do not point to existing files.
+        """
+        missing_files = [path for path in files if not path.is_file()]
```
nakul680, line 111: It would be good to resolve the file path here before checking whether the file exists. That could be done either here or before this method is called , I would leave that decision up to you .

nakul680 (changes requested):

PR #225, opened 2026-04-01
Version check added as Calculator attribute

crjacinto:
## Closes Issues
Closes #223
...

## Description
- Added a version check attribute to the Calculator class. When defined, it also overrides the version check of get_output. 

---

### Added 

- Add version check attribute to Calculator, which is parsed to get_output

### DIFF
--- a/src/opi/core.py
+++ b/src/opi/core.py
@@ -39,6 +39,9 @@ class Calculator:
         Can only be disabled after initialization of a `Calculator` (not recommended!).
     _input | input: Input
         Contains all ORCA input parameters except for the primary structural information.
+    version_check: bool, default: True
+        Activates version check when running an ORCA calculation. When running a complete series of input generation, calculation, 
+        and output processing, it overrides the get_output version check attribute.
     """
 
     def __init__(
@@ -94,6 +97,9 @@ class Calculator:
         # ----------------------------
         # > BINARY VERSION CHECK
         # ----------------------------
+
+        self.version_check: bool = version_check
+
         if version_check:
             # > Raises RuntimeError if version is not compatible or cannot be determined.
             self.check_version()
@@ -304,14 +310,19 @@ class Calculator:
         runner = self._create_runner()
         runner.create_jsons(self.basename, force=force)
 
-    def get_output(self) -> "Output":
+    def get_output(self, 
+                   *, 
+                   version_check: bool | None = None) -> "Output":
         """
         Get an instance of `Output` setup for the current job.
         Can be called before execution of job.
         """
+        vc = self.version_check if version_check is None else version_check
+
         return Output(
             basename=self.basename,
             working_dir=self.working_dir,
+            version_check=vc,
         )
 
     def check_version(self) -> None:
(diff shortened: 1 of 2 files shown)


src/opi/core.py
```diff
@@ -94,6 +97,9 @@ def __init__(
         # ----------------------------
         # > BINARY VERSION CHECK
         # ----------------------------
+
+        self.version_check: bool = version_check
+
         if version_check:
```
haneug: I would check for `self.version_check` here and use the attribute.

src/opi/core.py
```diff
@@ -39,6 +39,9 @@ class Calculator:
         Can only be disabled after initialization of a `Calculator` (not recommended!).
     _input | input: Input
         Contains all ORCA input parameters except for the primary structural information.
+    version_check: bool, default: True
+        Activates version check when running an ORCA calculation. When running a complete series of input generation, calculation,
```
haneug: I would try to stay closer to the description of the init parameter:
```
Check ORCA's binary version upon initialization. 
Will be handed to the initialization of Output in get_output.
```

src/opi/core.py
```diff
@@ -304,14 +310,17 @@ def create_jsons(self, *, force: bool = False) -> None:
         runner = self._create_runner()
         runner.create_jsons(self.basename, force=force)
 
-    def get_output(self) -> "Output":
+    def get_output(self, *, version_check: bool | None = None) -> "Output":
         """
         Get an instance of `Output` setup for the current job.
```
haneug, line 314: I would suggest to add a more descriptive docstring here and make clear that we have an optional keyword argument that overrides the class attribute.
```
Parameters
----------
version_check: bool | None, default = None
    Optional boolean that specifies whether a version check should be performed in the output.
    Overrides `self.version_check`.

```

haneug (changes requested):
Thank you for the contribution. I have some minor suggestions for improvement.

src/opi/core.py
timmyte: How about:
"Enable/disable ORCA binary version check as well as version check on the JSON output."?

src/opi/core.py
timmyte: 1) Correct to: **on the output**
2) The override behavior is not clear. What happens if the variable is `None`?

timmyte (commented):
Looks good to me. Some minor inconsistencies in the documentation ;)

src/opi/core.py
crjacinto: How about:

Parameters
----------
version_check : bool | None, default=None
    Whether to perform a version check on the output.
    If ``None``, the value of ``self.version_check`` is used.
    This argument overrides the instance attribute when provided.

src/opi/core.py
```diff
@@ -314,6 +313,12 @@ def get_output(self, *, version_check: bool | None = None) -> "Output":
...
         Can be called before execution of job.
+
+        Parameters
+        ----------
+        version_check : bool | None, default=None
+            Whether to perform a version check on the output.
+            If ``None``, the value of ``self.version_check`` is used.
```
timmyte:
```suggestion
 If `None`, the value of `self.version_check` is used.
```

timmyte (commented):
PR is good from my side.
The double backticks should be replaced with single ones.

haneug (approved):

haneug:
@crjacinto Thank you for contributing to OPI!

PR #226, opened 2026-04-02
Fix version check

haneug:
Small fix to version check.

### DIFF
--- a/src/opi/utils/orca_version.py
+++ b/src/opi/utils/orca_version.py
@@ -24,7 +24,7 @@ class OrcaVersion(Version):  # type: ignore
         re.VERBOSE | re.IGNORECASE,
     )
     RGX_OUTPUT_VERSION = re.compile(
-        r"^[ \t]*Program Version[ \t]+([1-9]+\.[0-9]+(\.[0-9]+(-f.[0-9]+)?)?)",
+        r"^[ \t]*Program Version[ \t]+([1-9]+\.[0-9]+(?:\.[0-9x]+)?(?:-f\.[0-9]+)?)",
         re.MULTILINE,
     )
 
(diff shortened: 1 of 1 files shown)


src/opi/utils/orca_version.py
```diff
@@ -24,7 +24,7 @@ class OrcaVersion(Version):  # type: ignore
         re.VERBOSE | re.IGNORECASE,
     )
     RGX_OUTPUT_VERSION = re.compile(
-        r"^[ \t]*Program Version[ \t]+([1-9]+\.[0-9]+(\.[0-9]+(-f.[0-9]+)?)?)",
+        r"^[ \t]*Program Version[ \t]+([1-9]+\.[0-9]+(?:\.[0-9x]+)?(?:-f\.[0-9]+)?)",
```
timmyte: Please change regex to:
```suggestion
r"^\s*Program Version\s+([1-9]+\.\d+\.([xX]|\d+(-f\.\d+)?))",
```
The non-capturing groups can be omitted, as we only use regex for fingerprinting.

src/opi/utils/orca_version.py
```diff
@@ -37,6 +37,10 @@ def from_str(cls, version_str: str, /) -> "OrcaVersion":
         ----------
         version_str : str
         """
+        # > Replace ".x" suffix with large number
+        if ".x" in version_str:
```
timmyte: Further down, we actually do dissect the version string into its components. Why not apply conversion there on the respective component?

timmyte (changes requested):
Great idea to fix this. I would how ever apply the fix slightly further down in the code on deeper level.

src/opi/utils/orca_version.py
haneug: I just commited this but your suggestions does not escape the dot in `-f.` and we do actually use the complete version string. Therefore I reverted it and would stay with the original one.

src/opi/utils/orca_version.py
haneug: To be honest I do not see an advantage in doing it explicitly since a x could show up in minor or micro/patch.

haneug:
Thanks I added your suggestions and also changed `RGX_VERSION` to allow the x in micro/path

timmyte (approved):
Looks good. Thanks for fixing this.

PR #229, opened 2026-04-15
Fixing server utilities

cplett:
## Description
- Adjusting the server functionality to the recent changes of the [oet](https://github.com/faccts/orca-external-tools). Previously, the server was always started with the python version used to execute OPI, which can lead to import errors when calling the `oet` scripts that were installed to a different environment.

---

### Fixed 
- Adjusted the `server` functionality to work with recent changes in the orca-external-tools infrastructure.

### DIFF
--- a/examples/exmp046_server/job.py
+++ b/examples/exmp046_server/job.py
@@ -28,8 +28,8 @@ server = OpiServer(
 
 # Start server. Change the method to the method that should be used, e.g., AIMNet2 or UMA.
 status = server.start_server(cmd_arguments="method")
-if status != ServerStatus.RUNNING:
-    if status == ServerStatus.ALREADY_RUNNING:
+if status is not ServerStatus.RUNNING:
+    if status is ServerStatus.ALREADY_RUNNING:
         warnings.warn("Some server was already running. Using old one.")
     elif status == ServerStatus.PORT_IN_USE:
         warnings.warn("Port for server in use. Continue on your own risk.")
--- a/examples/exmp046_server/job.py
+++ b/examples/exmp046_server/job.py
@@ -26,8 +26,8 @@ server = OpiServer(
     serverpath="/path/to/working/server/script/from/orca-external-tools"
 )
 
-# Start server
-status = server.start_server(exe="bash")
+# Start server. Change the method to the method that should be used, e.g., AIMNet2 or UMA.
+status = server.start_server(cmd_arguments="method")
 if status != ServerStatus.RUNNING:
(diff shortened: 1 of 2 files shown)


src/opi/external_methods/server.py
```diff
@@ -84,7 +84,7 @@ def _wait_for_port(self, host: str, port: int, timeout: float = 10.0) -> bool:
     def start_server(
         self,
         cmd_arguments: str | None = None,
-        exe: str = sys.executable,
+        exe: str | None = None,
```
timmyte, line 87: I would just use an empty string by default and avoid usage of `None` as nothing is gained from it.

examples/exmp046_server/job.py
```diff
@@ -26,8 +26,8 @@
     serverpath="/path/to/working/server/script/from/orca-external-tools"
 )
 
-# Start server
-status = server.start_server(exe="bash")
+# Start server. Change the method to the method that should be used, e.g., AIMNet2 or UMA.
+status = server.start_server(cmd_arguments="method")
 if status != ServerStatus.RUNNING:
     if status == ServerStatus.ALREADY_RUNNING:
```
timmyte:
```suggestion
if status is ServerStatus.RUNNING:
    if status is not ServerStatus.ALREADY_RUNNING:
```

src/opi/external_methods/server.py
```diff
@@ -95,7 +95,7 @@ def start_server(
         ----------
         cmd_arguments: str | None, default: None
             cmd arguments that should be passed to the server
-        exe: str, default: sys.executable
+        exe: str | None, default: None
```
timmyte, line 98: See comment above.

src/opi/external_methods/server.py
```diff
@@ -111,8 +111,8 @@ def start_server(
             # Start server by running a python process
             # Therefore, first set up the command line call for the server script
             # Build the command list:
-            # ["python", server_script] + -b ID:port + optional args
-            cmd = [exe, self.serverpath]
+            # optional exe + [server_script] + -b ID:port + optional args
+            cmd = [self.serverpath] if not exe else [exe, self.serverpath]
             cmd.append("-b")
             cmd.append(f"{self._host_id}:{self._port}")
```
timmyte: How about:
```suggestion
            cmd = [exe] if exe else []
            cmd += [
                self.serverpath,
                "-b",
                f"{self._host_id}:{self._port}",
            ]
```

timmyte (approved):
Thanks for fixing this. Please consider my remarks before merging. And also rerun Nox again.

PR #230, opened 2026-04-17
RMSD and Rotational Constants Modules 

crjacinto:
## Closes Issues
Closes #220 #219 

## Description
- Class methods to calculate the RMSD between two structures  and the rotational spectrum (rotor type, moments of inertia and rotational constants) of a molecule.

### Added 

- Added `rmsd` and `rmsd_kabsch` to calculate RMSD without and with rotational alignment to `Structure` class (#230). 
- Added `calc_rotational_constants` to calculate molecular rotational constants to `Structure` class (#230).
- Added `calc_rotor_type` to classify a molecule's rotor type to `Structure` class (#230).


src/opi/utils/rmsd.py
```diff
@@ -0,0 +1,108 @@
+import os
+from typing import List, Tuple, Union
+
+import numpy as np
+
+
+def read_xyz(data: Union[str, Tuple[List[str], np.ndarray]]) -> Tuple[List[str], np.ndarray]:
```
timmyte: 1) OPI has it's own `Structure` class. Please use that: https://github.com/faccts/opi/blob/fb149b9c0ffbbdd9821d892b4ccb5ebeca26bea0/src/opi/input/structures/structure.py
2) Please use following for numpy arrays in type annotation: `np.typing.NDArray[float]`

src/opi/utils/rmsd.py
```diff
@@ -0,0 +1,108 @@
...
+            symbols.append(sym)
+            coords.append(xyz)
+
+        return symbols, np.array(coords, dtype=np.float64)
+
+    else:
+        raise TypeError("Input must be: file path or XYZ string")
+
+
+def _validate_geometries(symA: List[str], symB: List[str]) -> None:
```
timmyte: Good idea to add such a function, but:
1) please add a docstring to every function.
2) stick to Python naming rules for variables and functions: https://realpython.com/python-variables/#rules-for-naming-variables

src/opi/utils/rmsd.py
```diff
@@ -0,0 +1,108 @@
...
+    if len(symA) != len(symB):
+        raise ValueError("Geometries have different number of atoms")
+
+    for i, (a, b) in enumerate(zip(symA, symB)):
+        if a != b:
+            raise ValueError(f"Atom mismatch at index {i}: {a} != {b}")
+
+
+def kabsch_rmsd(
+    ref_xyz: str,
```
timmyte: `ref_xyz` and `target_xyz` should be of type `Structure`.

src/opi/utils/rmsd.py
```diff
@@ -0,0 +1,108 @@
...
+    target_xyz: str,
+    *,
+    align: bool = True,
+) -> float:
+    """Compute RMSD between two XYZ geometries.
```
timmyte: The first text line of docstring should be a line of its own:

```suggestion
    """
    Compute RMSD between two XYZ geometries.
```

src/opi/utils/rmsd.py
```diff
@@ -0,0 +1,108 @@
...
+        RMSD (Å)
+    """
+
+    symA, A = read_xyz(ref_xyz)
+    symB, B = read_xyz(target_xyz)
+
+    _validate_geometries(symA, symB)
+
+    # Center using centroid (simple average)
+    A_cent = A - A.mean(axis=0)
```
timmyte: 1) When you start introducing `Structure` don't use the whole structure from here on. Extract the coordinates as Numpy array and only work with these, this is to make sure the structure are not modified.
2) Getting the coordinates of a `Structure` moved to the COM, would be nice object method `Structure.get_coordinates_com()`.

src/opi/utils/rmsd.py
```diff
@@ -0,0 +1,108 @@
...
+    U, _, Vt = np.linalg.svd(H)
+
+    d = np.linalg.det(Vt.T @ U.T)
+    D = np.diag([1.0, 1.0, d])
+
+    R = Vt.T @ D @ U.T
+
+    B_rot = B_cent @ R
+
+    diff = A_cent - B_rot
```
timmyte: I would outsource the plain-RMSD calculation into it's own function `plain_rmsd()`. Thereby you would also loose the `align` parameter, which for this function we would always assume to be True.

src/opi/utils/rmsd.py
```diff
@@ -0,0 +1,108 @@
...
+    B_cent = B - B.mean(axis=0)
+
+    if not align:
+        diff = A_cent - B_cent
+        return float(np.sqrt(np.sum(diff**2) / len(A_cent)))
+
+    # Standard Kabsch covariance matrix
+    H = B_cent.T @ A_cent
+
+    U, _, Vt = np.linalg.svd(H)
```
timmyte: Please add some comments to the individual operation, so one can follow what your are doing.

src/opi/utils/rotconst.py
```diff
@@ -0,0 +1,383 @@
...
+import warnings
+from dataclasses import dataclass
+from typing import List, Tuple, Union
+
+import numpy as np
+
+# ============================================================
+# Physical constants
+# ============================================================
+_AMU_TO_KG = 1.66053906660e-27
```
timmyte: Please move conversion factor to the appropriate module: https://github.com/faccts/opi/blob/main/src/opi/utils/units.py

src/opi/utils/rotconst.py
```diff
@@ -0,0 +1,383 @@
...
+_ANGSTROM_TO_M = 1.0e-10
+_H_PLANCK = 6.62607015e-34
+_C_CM = 2.99792458e10
+
+
+# ============================================================
+# Atomic masses
+# ============================================================
+ATOMIC_MASSES = {
```
timmyte: These should live here: https://github.com/faccts/opi/blob/main/src/opi/utils/element.py

src/opi/utils/rmsd.py
```diff
@@ -0,0 +1,108 @@
+import os
+from typing import List, Tuple, Union
```
timmyte: These imports are old-school. Please use modern-style type annotation:
* `List[int]` -> `list[int]`
* `tuple[int]` -> `tuple[int]`
* `Union[int, float]` -> `int | float`

src/opi/utils/rotconst.py
```diff
@@ -0,0 +1,383 @@
...
+    Flexible rotational constant calculator.
+
+    Input options
+    -------------
+    - symbols + coords
+    - xyz (file path, string, or lines)
+
+    Mass priority
+    -------------
+    masses > atom_weights > weights > default
```
timmyte: Fascinating. Is there a reason weigh we need 4 different weighting schemes?

src/opi/utils/rotconst.py
```diff
@@ -0,0 +1,383 @@
...
+    A = _moment_to_mhz(Ia)
+    B = _moment_to_mhz(Ib)
+    C = _moment_to_mhz(Ic)
+
+    # --- Rotor classification ---
+    tol = 1e-3
+    n_zero = sum(m < 1e-6 for m in (Ia, Ib, Ic))
+
+    if n_zero == 3:
+        rotor = "monoatomic"
```
timmyte: Nice idea to also determine the rotor type, but please don't use literal string constants. Have a look into enumerations, especially in this case Python's `StrEnum`.

src/opi/utils/rotconst.py
```diff
@@ -0,0 +1,383 @@
...
+    "Og": 294.0000,
+}
+
+
+# ============================================================
+# Dataclass
+# ============================================================
+@dataclass
+class RotationalConstants:
+    A: float | None
```
timmyte: Nice idea to encapsulate all this into a single class, but I would put into a different one.
I would add this to `Structure` and each quantity has it's own method:
```
Structure.calc_rotational_constants() -> np.array[float, float, float]
# First array are principle axis and the second the corresponding moments, sorted in ascending order by size.
Structure.calc_moment_of_inertia() -> np.array[float, float, float], -> np.array[float, float, float]
Structure.calc_rotor_type()  -> RotorType
```

timmyte (changes requested):
Solid first PR.
But the tools are bit to decoupled from OPI. You should use and add to the data structures that we already have in OPI.

timmyte:
As discussed, I would design the RMSD function as follows:
```python
class Structure:
    def rmsd(self, other: Structure, /, only_atoms: Sequence[int] = (), *, ignore_hs: bool = False) -> float:
        return _rmsd(self.coordinates, other.coordinates, only_atoms=only_atoms, ignore_hs=ignore_hs)

    def _rmsd(coords1: NDArray[float], coords2: NDArray[float], /, only_atoms: Sequence[int] = (), *, ignore_hs: bool = False)
        # We ignore `ignore_hs` if `only_atoms` is populated. This should also be documented
        if coords1.shape != coords2.shape:
            raise ValueError

    def rmsd_kabsch(self, other: Structure, /, only_atoms: Sequence[int] = (), *, ignore_hs: bool = False) -> float:
        # 1) Translate to center of geometry (centroid)
        # 2) find optimal rotation

        # `only_atoms` and `ignore_hs` don't need to be passed here anymore as `coords1` and `coords2_rotated` should only contain the coordinates of the relevant atoms.
        return _rmsd(coords1, coords2_rotated)
```

src/opi/utils/rmsd.py
crjacinto: Implemented through `Structure.centered_structure()`

src/opi/utils/rmsd.py
crjacinto: Implemented through `rmsd()` and `rmsd_kabsch()`

src/opi/utils/rotconst.py
crjacinto: Created module `constatns.py`

src/opi/utils/rotconst.py
crjacinto: To give the user different levels of control from specific to general. `masses` requires each atom mass specified by the user, `atom_weights` overrides atom weights by index, `weights` considers all atom weights of  a specific element, and `default` is the fallback to the `ATOMIC_MASSES` dict

src/opi/utils/rotconst.py
crjacinto: Now there are three different functions in `Structure` so one can request separately the Rotational Constants, the Principal Moments of Inertia, and the Rotor Type

src/opi/input/structures/structure.py
```diff
@@ -1066,3 +1074,363 @@ def _iter_xyz_structures(
...
+    def centered_structure(self) -> "Structure":
+        """
+        Return a new Structure centered at its centroid.
+
+        Returns
+        -------
+        Structure
+            New Structure with centered coordinates.
+        """
+        import copy
```
timmyte: Don't do local imports. They should ALWAYS be done on module level.

src/opi/input/structures/structure.py
```diff
@@ -1066,3 +1074,363 @@ def _iter_xyz_structures(
...
+
+        coords = np.array([a.coordinates.coordinates for a in self.atoms], dtype=np.float64)
+        centered_coords = coords - coords.mean(axis=0)
+
+        new_structure = copy.deepcopy(self)
+        for atom, new_coord in zip(new_structure.atoms, centered_coords):
```
timmyte: This should be facilicated by a method of `Structure` -> `Structure.set_coordinates()`

src/opi/input/structures/structure.py
```diff
@@ -1066,3 +1074,363 @@ def _iter_xyz_structures(
...
+        Return a new Structure centered at its centroid.
+
+        Returns
+        -------
+        Structure
+            New Structure with centered coordinates.
+        """
+        import copy
+
+        coords = np.array([a.coordinates.coordinates for a in self.atoms], dtype=np.float64)
```
timmyte: `Structure` should have method for this -> `Structure.get_coordinates()`

src/opi/input/structures/structure.py
```diff
@@ -1066,3 +1074,363 @@ def _iter_xyz_structures(
...
+        ignore_hs: bool,
+    ) -> list[Atom]:
+        """
+        Return Atom instances after applying only_atoms and ignore_hs filters.
+        """
+        atom_list = self.atoms
+        if only_atoms:
+            candidates = [atom_list[i] for i in only_atoms]
+        elif ignore_hs:
+            candidates = [a for a in atom_list if isinstance(a, Atom) and a.element != Element.H]
```
timmyte: Please document, why would have to check for `isinstance(a, Atom)` here.

src/opi/input/structures/structure.py
```diff
@@ -1066,3 +1074,363 @@ def _iter_xyz_structures(
...
+        else:
+            candidates = [a for a in atom_list if isinstance(a, Atom)]
```
timmyte: Please document, why would have to check for `isinstance(a, Atom)` here.

src/opi/input/structures/structure.py
```diff
@@ -1066,3 +1074,363 @@ def _iter_xyz_structures(
...
+        """
+        Raise ValueError if atoms1 and atoms2 differ in count or element order.
+        """
+        if len(atoms1) != len(atoms2):
+            raise ValueError(
+                f"Structures have different number of atoms: {len(atoms1)} vs {len(atoms2)}"
+            )
+        for i, (a, b) in enumerate(zip(atoms1, atoms2)):
+            if a.element != b.element:
+                raise ValueError(f"Atom mismatch at index {i}: {a.element!r} != {b.element!r}")
```
timmyte: Use natural counting (starting from 1). Humans don't count from zero.
You can use `enumerate(start=1)` for this.

src/opi/input/structures/structure.py
```diff
@@ -1066,3 +1074,363 @@ def _iter_xyz_structures(
...
+        """
+        Compute RMSD between this structure and *other* (no rotational alignment).
+
+        Both structures are translated to their centroid before comparison.
+
+        Parameters
+        ----------
+        other : Structure
+            Structure to compare against.
+        only_atoms : Sequence[int], default ()
```
timmyte: `default:`

src/opi/input/structures/structure.py
```diff
@@ -1066,3 +1074,363 @@ def _iter_xyz_structures(
...
+            Atom indices to include. If non-empty, ``ignore_hs`` is ignored.
+        ignore_hs : bool, default False
```
timmyte: `default:`

src/opi/input/structures/structure.py
```diff
@@ -1066,3 +1074,363 @@ def _iter_xyz_structures(
...
+            If the filtered atom sets differ in size or element order.
+        """
+        atoms1 = self._filtered_atoms(only_atoms, ignore_hs)
+        atoms2 = other._filtered_atoms(only_atoms, ignore_hs)
+        self._validate_rmsd_compatibility(atoms1, atoms2)
+
+        coords1 = np.array([a.coordinates.coordinates for a in atoms1], dtype=np.float64)
+        coords2 = np.array([a.coordinates.coordinates for a in atoms2], dtype=np.float64)
+
+        coords1 -= coords1.mean(axis=0)
```
timmyte: I would add method for this -> `Structure.get_coordinates_at_centroid()`.

src/opi/input/structures/structure.py
```diff
@@ -1066,3 +1074,363 @@ def _iter_xyz_structures(
...
+    def calc_moment_of_inertia(
+        self,
+        masses: npt.NDArray[np.float64] | None = None,
+        weights: dict[str, float] | None = None,
+        atom_weights: dict[int, float] | None = None,
+    ) -> PrincipalMoments | None:
+        """
+        Compute the principal axes and moments of inertia for this structure.
+
+        Only ``Atom`` instances contribute; ``GhostAtom``, ``PointCharge``,
```
timmyte: Single back-ticks are fine.

src/opi/input/structures/structure.py
```diff
@@ -1066,3 +1074,363 @@ def _iter_xyz_structures(
...
+            other mass source.
+        atom_weights : dict[int, float] | None, default None
+            Per-atom mass overrides keyed by index within the Atom list.
+        weights : dict[str, float] | None, default None
+            Per-element mass overrides keyed by element symbol string
+            (e.g. ``{"C": 13.003}``).
+
+        Returns
+        -------
+        tuple[ndarray shape (3, 3), ndarray shape (3,)] | None
```
timmyte: The return type looks incorrect to me! Should this be `PrincipalMoments`?

src/opi/input/structures/structure.py
```diff
@@ -1066,3 +1074,363 @@ def _iter_xyz_structures(
...
+        coords2_rotated = coords2_centered @ R
+
+        return self._rmsd_coords(coords1, coords2_rotated)
+
+    # ------------------------------------------------------------------ #
+    #  Moment of inertia                                                   #
+    # ------------------------------------------------------------------ #
+    def calc_moment_of_inertia(
+        self,
+        masses: npt.NDArray[np.float64] | None = None,
```
timmyte: Fancy idea to have all this mass parameters, but it does bloat the API quite.

@haneug 
What's your take?

src/opi/input/structures/structure.py
```diff
@@ -1066,3 +1074,363 @@ def _iter_xyz_structures(
...
+        R = Vt.T @ D @ U.T
+
+        coords2_rotated = coords2_centered @ R
+
+        return self._rmsd_coords(coords1, coords2_rotated)
+
+    # ------------------------------------------------------------------ #
+    #  Moment of inertia                                                   #
+    # ------------------------------------------------------------------ #
+    def calc_moment_of_inertia(
```
timmyte: I would go with the plural: `calc_moments_of_inertia()`

src/opi/utils/rotconst.py
```diff
@@ -0,0 +1,163 @@
...
+
+
+def mhz_to_wavenumber(mhz: float | None) -> float | None:
+    """Convert a rotational constant from MHz to cm⁻¹."""
+    if mhz is None:
+        return None
+    return mhz * 1e6 / constants.C
+
+
+def classify_rotor_type(moments: np.ndarray, tol: float = 1e-3) -> RotorType:
```
timmyte: I would turn this into an instance method of `PrincipalMoments`. Also there is no need to store the output anywhere.

src/opi/utils/rotconst.py
```diff
@@ -0,0 +1,163 @@
...
+# ============================================================
+
+
+def moment_to_mhz(inertia: float) -> float | None:
+    """
+    Convert a principal moment of inertia (amu·Å²) to a rotational
+    constant in MHz.  Returns None when the moment is effectively zero
+    (degenerate / linear axis).
+    """
+    if inertia < 1e-6:
```
timmyte: Why can `inertia` not be `None`, but `mhz` in the method below.
Please align the APIs of these functions.

src/opi/utils/rotconst.py
```diff
@@ -0,0 +1,163 @@
...
+    """
+    Stores rotational constants and molecular moments of inertia.
+
+    Attributes
+    ----------
+    A, B, C : float | None
+        Rotational constants in MHz (None for a degenerate axis).
+    A_cm, B_cm, C_cm : float | None
+        Rotational constants in cm⁻¹.
+    moments : tuple[float, float, float]
```
timmyte: Please double check.
This class does neither have a `moments` nor `rotor_type`.
But see my suggestion at the bottom about having a method to determine the `rotr_type` on the fly.

src/opi/utils/rotconst.py
```diff
@@ -0,0 +1,163 @@
...
+    rotor_type : RotorType
+        Molecular rotor classification.
+    """
+
+    def __init__(
+        self,
+        A: float | None,
+        B: float | None,
+        C: float | None,
+        A_cm: float | None,
```
timmyte: Don't store directly related values in the same object. The object would be responsible to keep them in sync.

If I do this
```
rcs = RotationalConstants(A=10, A_cm=10*CONVERSION_FACTOR)
rcs.A *= 5
```
Then `rcs.A` and `rcs.A_cm` are out-of-sync. If you want to offer simple API to get the values of the constants in different unit, just add method that returns in that unit -> `RotationalConstants.get_in_wavenumbers()` which return a tuple with the three values.

src/opi/input/structures/structure.py
```diff
@@ -1066,3 +1074,363 @@ def _iter_xyz_structures(
...
+            If the length of *masses* does not match the number of ``Atom``
+            instances in the structure.
+        """
+        atom_list = [a for a in self.atoms if isinstance(a, Atom)]
+        if not atom_list:
+            return None
+
+        coords = np.array([a.coordinates.coordinates for a in atom_list], dtype=np.float64)
+
+        weights = {k: v for k, v in (weights or {}).items()}
```
timmyte: Please add some documentation here to these block where you sort our which mass or weights to take.

src/opi/input/structures/structure.py
```diff
@@ -1066,3 +1074,363 @@ def _iter_xyz_structures(
...
+                masses_list.append(m)
+            masses = np.array(masses_list, dtype=np.float64)
+
+        mask = masses > 0.0
+        if not np.any(mask):
+            return None
+        masses = masses[mask]
+        coords = coords[mask]
+
+        total_mass = float(masses.sum())
```
timmyte, line 1285: Also here. Add brief comments before each block, documenting what the block does.

src/opi/input/structures/structure.py
```diff
@@ -1066,3 +1074,363 @@ def _iter_xyz_structures(
...
+        weights: dict[str, float] | None = None,
+        atom_weights: dict[int, float] | None = None,
+    ) -> PrincipalMoments | None:
+        """
+        Compute the principal axes and moments of inertia for this structure.
+
+        Only ``Atom`` instances contribute; ``GhostAtom``, ``PointCharge``,
+        and ``EmbeddingPotential`` atoms are silently ignored.
+
+        Mass priority
```
timmyte: This block should come after `Parameters`. This applies to every docstring that has this.

First explain your terminology, then you can explain the rules.

src/opi/input/structures/structure.py
```diff
@@ -1066,3 +1074,363 @@ def _iter_xyz_structures(
...
+        RotorType | None
+            ``None`` if the structure has no real atoms or all masses vanish.
+        """
+        if moments is None:
+            result = self.calc_moment_of_inertia(**mass_kwargs)
+            if result is None:
+                return None
+            moments = result
+
+        return classify_rotor_type(np.array([moments.Ia, moments.Ib, moments.Ic]))
```
timmyte: See my suggestion at the definition of `classify_rotor_type()`.

timmyte (changes requested):
Very nice improvement of the code. We are getting there.
I really appreciate how well you documented the code.

src/opi/input/structures/structure.py
crjacinto: Added method set_coordinates()

src/opi/input/structures/structure.py
crjacinto: Added documentation. This checks that explicitly indexed atom positions doesn't introduce GhostAtoms, PointCharges, etc. However, I changed it to type(a) is Atom since GhostAtom seems to be a subclass of Atom

src/opi/input/structures/structure.py
crjacinto: Added method get_coordinates().

src/opi/input/structures/structure.py
crjacinto: Same as before

src/opi/input/structures/structure.py
crjacinto: Done, thanks

src/opi/input/structures/structure.py
crjacinto: Added the method

src/opi/input/structures/structure.py
crjacinto: Changed hopefully everywhere

src/opi/input/structures/structure.py
crjacinto: You're right. I changed the code but not the documentation. Thank you.

src/opi/input/structures/structure.py
crjacinto: Done

src/opi/utils/rotconst.py
crjacinto: Changed both, thanks

src/opi/utils/rotconst.py
crjacinto: Everything aligned now.

src/opi/utils/rotconst.py
crjacinto: Done, thanks for clarifying

src/opi/input/structures/structure.py
crjacinto: Change made, thanks for the explanation

src/opi/utils/rotconst.py
crjacinto: Same error of changing the code but not the documentation. Now solved, thanks

src/opi/input/structures/structure.py
```diff
@@ -1066,3 +1073,423 @@ def _iter_xyz_structures(
             n_struc += 1
             if n_struc_limit and n_struc >= n_struc_limit:
                 break
+
+    def get_coordinates(
+        self,
+        atoms: Sequence[Atom | EmbeddingPotential | GhostAtom | PointCharge] | None = None,
```
timmyte: Great. Thanks for adding the routine!
Looks good, but I don't get why you would pass it a list of `Atom`-like objects.
If you already know `Atom`s, you can just loop over them and collect their coordinates, making this routine superfluous.

If you like to be able to pass a list to limit the atoms, I would go with their position index in `self.atoms`, rather than the objects themselves.

src/opi/input/structures/structure.py
```diff
@@ -1066,3 +1073,423 @@ def _iter_xyz_structures(
...
+        ValueError
+            If *coords* shape does not match the number of atoms.
+        """
+
+        coords = np.asarray(coords, dtype=np.float64)
+        if coords.shape != (len(self.atoms), 3):
+            raise ValueError(
+                f"coords shape {coords.shape} does not match expected ({len(self.atoms)}, 3)"
+            )
+        new_structure = copy.deepcopy(self)
```
timmyte: A setter usually updates the current object.
Therefore, I would act on the current object directly. I would consider this the more flexible solution, as the decision to work with a copy stays with the user.

src/opi/input/structures/structure.py
```diff
@@ -1066,3 +1073,423 @@ def _iter_xyz_structures(
...
+        `GhostAtom`, `PointCharge`, and `EmbeddingPotential` are ignored.
+
+        Returns
+        -------
+        Structure
+            New `Structure` with centered coordinates.
+        """
+        real_atoms = [a for a in self.atoms if type(a) is Atom]
+        centroid = self.get_coordinates(atoms=real_atoms).mean(axis=0)
+        return self.set_coordinates(self.get_coordinates() - centroid)
```
timmyte: I would make the deepcopy here instead.

src/opi/input/structures/structure.py
```diff
@@ -1066,3 +1073,423 @@ def _iter_xyz_structures(
...
+        only_atoms: Sequence[int],
+        ignore_hs: bool,
+    ) -> list[Atom]:
+        """
+        Return real `Atom` instances after applying `only_atoms` and `ignore_hs` filters.
+
+        The final `type(a) is Atom` check ensures that explicitly indexed atoms
+        (via `only_atoms`) cannot introduce `GhostAtom`, `PointCharge`, or
+        `EmbeddingPotential` instances into the returned list. Note that `GhostAtom`
+        is a subclass of `Atom`, so `isinstance` is intentionally avoided here.
```
timmyte: Very good!

src/opi/input/structures/structure.py
```diff
@@ -1066,3 +1073,423 @@ def _iter_xyz_structures(
...
+            The atoms to extract coordinates from. If `None`, all entries
+            in `self.atoms` are used.
+
+        Returns
+        -------
+        npt.NDArray[np.float64], shape (N, 3)
+            Coordinates in the same order as the input list.
+        """
+        atom_list = atoms if atoms is not None else self.atoms
+        return np.array([a.coordinates.coordinates for a in atom_list], dtype=np.float64)
```
timmyte: We should refrain from setting `dtype` explicitly everywhere, as this introduces an extra chore, which might be overlooked by developers.

Or is there a particular reason, why we need to set this explicitly?

src/opi/input/structures/structure.py
```diff
@@ -1066,3 +1073,423 @@ def _iter_xyz_structures(
...
+            Centroid-translated coordinates in the same order as the input list.
+        """
+        coords = self.get_coordinates(atoms=atoms)
+        return coords - np.mean(coords, axis=0, dtype=np.float64)
+
+    def rmsd(
+        self,
+        other: "Structure",
+        /,
+        only_atoms: Sequence[int] = (),
```
timmyte, line 1121: See, here a list of indices is passed, where above in `get_coordinates()` you used a list of `Atom` objects.
Also take into account that APIs should be consistent (if possible) across different routines in a project.

src/opi/input/structures/structure.py
timmyte: Did you forget to push the change?

tests/unit/test_rmsd.py
```diff
@@ -0,0 +1,430 @@
...
+
+# ============================================================
+# Fixtures
+# ============================================================
+
+
+@pytest.fixture()
+def water() -> Structure:
+    """H2O at a known geometry."""
+    return make_structure(
```
timmyte: To simplify the API a little, I would imply `make_atom()` in `make_structure()`, i.e.:
```
return make_structure([
   ("O", 0.000000, 0.000000, 0.119748),
   ("H", 0.000000, 0.756950, -0.478993),
   ("H", 0.000000, -0.756950, -0.478993),
])
```

timmyte (commented):
Thanks for tackling my remarks. There are still some unresolved issues though.

Please also see the few additional points I raised.

src/opi/input/structures/structure.py
```diff
@@ -1066,3 +1073,430 @@ def _iter_xyz_structures(
             n_struc += 1
             if n_struc_limit and n_struc >= n_struc_limit:
                 break
+
+    @property
+    def real_atoms(self) -> list[Atom]:
+        """
```
timmyte: Good idea to add this function!!

src/opi/input/structures/structure.py
timmyte: @haneug 
Any comment?

src/opi/input/structures/structure.py
```diff
@@ -1066,3 +1073,430 @@ def _iter_xyz_structures(
...
+
+        if masses is not None:
+            masses = np.asarray(masses, dtype=np.float64)
+            if len(masses) != len(self.real_atoms):
+                raise ValueError(
+                    f"masses length ({len(masses)}) does not match "
+                    f"number of atoms ({len(self.real_atoms)})"
+                )
+        else:
+            masses_list: list[float] = []
```
timmyte: Please add some comments to all these if statements. What does each do!?

src/opi/utils/rotconst.py
```diff
@@ -0,0 +1,202 @@
...
+    axes : np.ndarray, shape (3, 3)
+        Corresponding eigenvectors (columns).
+    """
+
+    Ia: float
+    Ib: float
+    Ic: float
+    axes: np.ndarray
+
+    # def rotor_type(self, tol: float = 1e-3) -> RotorType:
```
timmyte: This can be removed, I assume.

timmyte (approved):
Looks good. Please consider my last remark before merging.
I also would like to get @haneug opinion on the API with all the different weight parameters

src/opi/input/structures/structure.py
haneug: Sorry for replying so delayed: I agree that I find the API here a bit bloated. I would just take whatever mass is in the atom and if None was set we use the default. The element-wise weights `weights` I find useful since the input can be given by hand but `masses` and `atom_weights` I find a bit too much. I would suggest to rename `weights` to `elem_masses`.

haneug: Or leave the element masses out as well and add a function that allows to easily set the mass for all atoms of a specific element.

src/opi/input/structures/structure.py
```diff
@@ -1066,3 +1073,425 @@ def _iter_xyz_structures(
...
+        other: "Structure",
+        /,
+        only_atoms: Sequence[int] = (),
+        *,
+        ignore_hs: bool = False,
+    ) -> float:
+        """
+        Compute RMSD between this structure and *other* (no rotational alignment).
+
+        Both structures are translated to their centroid before comparison.
```
timmyte, line 1128: I would be a bit more precise here:
The two structure objects themselves are not modified. Please mention this.

If the objects themselves were changed, there could be side-effects in any code coming after calling this method. This is important to avoid!

src/opi/input/structures/structure.py
```diff
@@ -1066,3 +1073,425 @@ def _iter_xyz_structures(
...
+        other: "Structure",
+        /,
+        only_atoms: Sequence[int] = (),
+        *,
+        ignore_hs: bool = False,
+    ) -> float:
+        """
+        Compute RMSD between this structure and *other* using the Kabsch algorithm.
+
+        Translates both structures to their centroid, then finds the optimal
```
timmyte, line 1173: Same here. Mention that the `Structure` object themselves are not modified.

src/opi/input/structures/structure.py
```diff
@@ -1066,3 +1073,425 @@ def _iter_xyz_structures(
...
+        R = Vt.T @ D @ U.T
+
+        return self._rmsd_coords(coords1, coords2 @ R)
+
+    # ------------------------------------------------------------------ #
+    #  Moment of inertia                                                   #
+    # ------------------------------------------------------------------ #
+    def _resolve_masses(
+        self,
+        elem_masses: dict[str, float] | None = None,
```
```suggestion
        elem_masses: dict[str | Element, float] | None = None,
```

src/opi/input/structures/structure.py
```diff
@@ -1066,3 +1073,425 @@ def _iter_xyz_structures(
...
+            else:
+                # Unknown element — warn and exclude from computation
+                warn(f"Unknown element '{atom.element.value}' → mass set to 0.0")
+                m = 0.0
+            masses_list.append(m)
+        return np.array(masses_list, dtype=np.float64)
+
+    def calc_moments_of_inertia(
+        self,
+        elem_masses: dict[str, float] | None = None,
```
```suggestion
        elem_masses: dict[str | Element, float] | None = None,
```

src/opi/input/structures/structure.py
```diff
@@ -1066,3 +1073,425 @@ def _iter_xyz_structures(
...
+        Classify the molecular rotor type.
+
+        Parameters
+        ----------
+        moments : PrincipalMoments | None, default None
+            Pre-computed principal moments (amu·Å², ascending). When
+            `None` the moments are computed via
+            :meth:`calc_moments_of_inertia` using *mass_kwargs*.
+        **mass_kwargs
+            Forwarded to :meth:`calc_moments_of_inertia` when *moments* is
```
```suggestion
            Forwarded to :meth:`calc_moments_of_inertia()` when `moments` is
```

src/opi/input/structures/structure.py
```diff
@@ -1066,3 +1073,425 @@ def _iter_xyz_structures(
...
+        if moments is None:
+            result = self.calc_moments_of_inertia(**mass_kwargs)
+            if result is None:
+                return None
+            moments = result
+        return moments.rotor_type()
+
+    def calc_rotational_constants(
+        self,
+        elem_masses: dict[str, float] | None = None,
```
```suggestion
        elem_masses: dict[str | Element, float] | None = None,
```

timmyte: Also update the docstring below, please.

src/opi/input/structures/structure.py
timmyte: Also update the docstring below, please.

src/opi/input/structures/structure.py
timmyte: Also update the docstring below, please.

src/opi/input/structures/structure.py
```diff
@@ -1066,3 +1073,425 @@ def _iter_xyz_structures(
...
+    ) -> RotationalConstants | None:
+        """
+        Compute rotational constants for this structure.
+
+        Only `Atom` instances contribute; `GhostAtom`, `PointCharge`,
+        and `EmbeddingPotential` atoms are silently ignored.
+
+        Mass priority
+        -------------
+        elem_masses > default (ATOMIC_MASSES_FROM_ELEMENT)
```
timmyte: What about `atom.mass`, like in the other routines that have the `elem_masses` parameter?

src/opi/input/structures/structure.py
```diff
@@ -1066,3 +1073,425 @@ def _iter_xyz_structures(
...
+                m = elem_masses[atom.element.value]
+            elif atom.mass is not None:
+                # Mass set directly on the atom takes second priority
+                m = atom.mass
+            elif atom.element in ATOMIC_MASSES_FROM_ELEMENT:
+                # Fall back to default atomic masses
+                m = ATOMIC_MASSES_FROM_ELEMENT[atom.element]
+            else:
+                # Unknown element — warn and exclude from computation
+                warn(f"Unknown element '{atom.element.value}' → mass set to 0.0")
```
timmyte: This should raise an exception, rather than a warning.

Warnings are meant for situations, where the user should be informed about problems that do not (significantly) influence the outcome of the program.

However, ignoring parts of a molecule, that should be proper `Atom`s, definitely influences the method's outcome.

Also consider that warnings are typically not caught like exceptions. So a program cannot actively react to a warning.

src/opi/input/structures/structure.py
```diff
@@ -1066,3 +1073,425 @@ def _iter_xyz_structures(
...
+        Only `Atom` instances contribute; `GhostAtom`, `PointCharge`,
+        and `EmbeddingPotential` atoms are silently ignored.
+
+        Mass priority
+        -------------
+        elem_masses > default (ATOMIC_MASSES_FROM_ELEMENT)
+
+        Parameters
+        ----------
+        elem_masses : dict[str, float] | None, default None
```
```suggestion
        elem_masses : dict[str | Element, float] | None, default: None
```

src/opi/input/structures/structure.py
```diff
@@ -1066,3 +1073,425 @@ def _iter_xyz_structures(
...
+        """
+        Resolve masses for all real `Atom` instances without mutating any atom.
+
+        Mass priority
+        -------------
+        elem_masses > atom.mass > default (ATOMIC_MASSES_FROM_ELEMENT)
+
+        Parameters
+        ----------
+        elem_masses : dict[str, float] | None, default None
```
```suggestion
        elem_masses : dict[str | Element, float] | None, default: None
```

src/opi/input/structures/structure.py
```diff
@@ -1066,3 +1073,425 @@ def _iter_xyz_structures(
...
+        `GhostAtom` is a subclass of `Atom`, so an exact type check
+        (`type(a) is Atom`) is used rather than `isinstance`.
+
+        Mass priority
+        -------------
+        elem_masses > atom.mass > default (ATOMIC_MASSES_FROM_ELEMENT)
+
+        Parameters
+        ----------
+        elem_masses : dict[str, float] | None, default None
```
```suggestion
        elem_masses: dict[str | Element, float] | None, default: None
```

src/opi/input/structures/structure.py
```diff
@@ -1066,3 +1073,425 @@ def _iter_xyz_structures(
...
+        self,
+        moments: PrincipalMoments | None = None,
+        **mass_kwargs: Any,
+    ) -> RotorType | None:
+        """
+        Classify the molecular rotor type.
+
+        Parameters
+        ----------
+        moments : PrincipalMoments | None, default None
```
```suggestion
        moments : PrincipalMoments | None, default: None
```

src/opi/input/structures/structure.py
```diff
@@ -1066,3 +1073,425 @@ def _iter_xyz_structures(
...
+            Forwarded to :meth:`calc_moments_of_inertia` when *moments* is
+            `None` (i.e. `elem_masses`).
+
+        Returns
+        -------
+        RotorType | None
+            `None` if the structure has no real atoms or all masses vanish.
+        """
+        if moments is None:
+            result = self.calc_moments_of_inertia(**mass_kwargs)
```
timmyte: Why not set assign the return value directly to `moments` here?

src/opi/input/structures/structure.py
```diff
@@ -1066,3 +1073,425 @@ def _iter_xyz_structures(
...
+        total_mass = float(masses.sum())
+        com = (masses[:, None] * coords).sum(axis=0) / total_mass
+        coords -= com
+
+        inertia = np.zeros((3, 3), dtype=np.float64)
+        for m, r in zip(masses, coords):
+            inertia += m * (np.dot(r, r) * np.eye(3) - np.outer(r, r))
+
+        moments_raw, axes = np.linalg.eigh(inertia)  # ascending order guaranteed
+        moments_raw = np.maximum(moments_raw, 0.0)
```
timmyte, line 1297: I would have guessed the eigenvalues cannot be negative, so why the `0.0` here?

tests/unit/test_structure_rmsd.py
```diff
@@ -0,0 +1,476 @@
...
+from opi.input.structures.coordinates import Coordinates
+from opi.input.structures.structure import Structure
+from opi.utils.element import Element
+
+# ============================================================
+# Helpers
+# ============================================================
+
+
+def make_atom(element: str, x: float, y: float, z: float) -> Atom:
```
timmyte, line 34: Please add a docstring.

tests/unit/test_structure_rmsd.py
```diff
@@ -0,0 +1,476 @@
...
+    return Atom(
+        element=Element(element),
+        coordinates=Coordinates(coordinates=(x, y, z)),
+    )
+
+
+def make_structure(*atoms: Atom) -> Structure:
```
timmyte, line 41: Please add a docstring.

tests/unit/test_structure_rmsd.py
```diff
@@ -0,0 +1,476 @@
...
+    )
+
+
+@pytest.fixture()
+def water_rotated(water) -> Structure:
+    """
+    H2O rotated 90° around Z after centring.
+    Kabsch RMSD vs the original centred water should be ~0.
+    """
+    centered = water.centered_structure()
```
timmyte, line 78: Hard-code the coordinates.

If fixtures are based on functions they are supposed to test (or related functions), then this can result in masked bugs.

tests/unit/test_structure_rmsd.py
```diff
@@ -0,0 +1,476 @@
...
+
+    def test_empty_structure(self):
+        structure = Structure(
+            atoms=[PointCharge(coordinates=Coordinates(coordinates=(0.0, 0.0, 0.0)), charge=1.0)]
+        )
+        assert structure.real_atoms == []
+
+
+# ============================================================
+# get_coordinates
```
timmyte, line 165: When referring to routines add `()` to the end, to signal that this routine. This is just good practice ;)
Make sure to do this throughout any piece of code documentation: comments, docstring, etc.

tests/unit/test_structure_rmsd.py
```diff
@@ -0,0 +1,476 @@
...
+            ]
+        )
+        assert len(structure.real_atoms) == 1
+        assert type(structure.real_atoms[0]) is Atom
+
+    def test_all_real_atoms(self, water):
+        """All atoms in water are real."""
+        assert len(water.real_atoms) == 3
+
+    def test_return_type(self, water):
```
timmyte, line 153: Please add docstring, briefly describing the purpose of the test.

tests/unit/test_structure_rmsd.py
```diff
@@ -0,0 +1,476 @@
...
+
+
+# ============================================================
+# get_coordinates
+# ============================================================
+
+
+class TestGetCoordinates:
+    def test_shape(self, water):
+        coords = water.get_coordinates()
```
timmyte, line 173: Please add a docstring briefly describing the purpose of the test.

tests/unit/test_structure_rmsd.py
```diff
@@ -0,0 +1,476 @@
...
+
+# ============================================================
+# get_coordinates_at_centroid
+# ============================================================
+
+
+class TestGetCoordinatesAtCentroid:
+    def test_centroid_is_zero(self, water):
+        coords = water.get_coordinates_at_centroid()
+        np.testing.assert_allclose(coords.mean(axis=0), [0.0, 0.0, 0.0], atol=1e-12)
```
timmyte, line 217: Please document why so picked this value for `atol`?
Please do this for all tests below.
It's not obvious why its 1e-12 here and 1e-6 below.

tests/unit/test_structure_rmsd.py
```diff
@@ -0,0 +1,476 @@
...
+
+
+# ============================================================
+# set_coordinates
+# ============================================================
+
+
+class TestSetCoordinates:
+    def test_mutates_in_place(self, water):
+        """set_coordinates should modify the structure in place."""
```
timmyte, line 242:
```suggestion
        """set_coordinates should modify the structure in-place."""
```

tests/unit/test_rotconst.py
```diff
@@ -0,0 +1,481 @@
...
+        pm = water.calc_moments_of_inertia()
+        assert pm is not None
+        product = pm.axes.T @ pm.axes
+        np.testing.assert_allclose(product, np.eye(3), atol=1e-10)
+
+    def test_str_output(self, water):
+        pm = water.calc_moments_of_inertia()
+        assert pm is not None
+        s = str(pm)
+        assert "Moments of inertia (amu·Å²):" in s
```
timmyte: You should know the extract string that comes out.
Why not compare directly against that?

timmyte (changes requested):
Very good, especially the exhaustive amount of tests. BRAVO!

src/opi/input/structures/structure.py
timmyte: `dtype` is still set explicitly. Please remove this, if there is no particular reason to do so. Otherwise, add comment mentioning the reason.

src/opi/input/structures/structure.py
timmyte: Please move the block after `Parameters`.

src/opi/input/structures/structure.py
```diff
@@ -1475,13 +1494,14 @@ def calc_rotational_constants(
 
         Mass priority
```
timmyte, line 1345: Same here: Please make that the `Parameters` block comes first and then the `Mass  priority`

src/opi/utils/rotconst.py
```diff
@@ -170,7 +170,7 @@ def moment_to_mhz(inertia: float | None) -> float | None:
     constant in MHz. Returns `None` when the moment is `None` or
     effectively zero (degenerate / linear axis).
     """
-    if inertia is None or inertia < 1e-6:
+    if inertia is None or inertia < 1e-3:
```
timmyte, line 177: Why did this number change now!?
Please document why exactly `10e-3`.

tests/unit/test_rmsd.py
```diff
@@ -431,6 +435,7 @@ def test_translated_zero_rmsd(self, water, water_translated):
     def test_rotated_zero_rmsd(self, water, water_rotated):
         """Kabsch should align the rotation and give ~0 RMSD."""
         centered = water.centered_structure()
+        print(centered.get_coordinates())
```
timmyte: Why the print statement here?

timmyte (dismissed):
Thanks for quickly tackling my remarks. But please make sure you addressed all of them.

tests/unit/test_structure_rmsd.py
```diff
@@ -0,0 +1,480 @@
...
+        ]
+    )
+
+
+# ============================================================
+# real_atoms
+# ============================================================
+
+
+class TestRealAtoms:
```
haneug, line 130: Please add `@pytest.mark.unit` to all your tests since these are unit tests and should be included as such! You can add it on the class level.

src/opi/utils/constants.py
```diff
@@ -0,0 +1,8 @@
+"""
+Contains universal constants.
+"""
+
+# Planck's Constant [m^2*kg/s]
+H_PLANCK = 6.62607015e-34
+# Speed of light in vacuum [m/s]
```
haneug, line 7: This is in cm/s and not in m/s! In your implementation this seems to be fine but if someone else uses it they should be aware of the correct units!

tests/unit/test_rotconst.py
```diff
@@ -0,0 +1,483 @@
...
+
+    def test_elem_masses_priority_over_atom_mass(self, water):
+        """`elem_masses` should take priority over `atom.mass` for the same atom."""
+        water.real_atoms[0].mass = 17.999
+        pm_atom_mass = water.calc_moments_of_inertia()
+        pm_elem = water.calc_moments_of_inertia(elem_masses={"O": 16.0})
+        assert pm_atom_mass is not None and pm_elem is not None
+        assert pm_atom_mass.Ic != pytest.approx(pm_elem.Ic, rel=1e-3)
+
+    def test_unknown_element_warns_and_excludes(self):
```
haneug: I think this test is not really necessary. If an unknown element without a mass shows up a value error is raised (which will not happen now since every enum member has a mass, it is done to be defensive). I do not see the point here what is exactly tested since C and H are both known.

haneug (changes requested):
Please address all of the comments by Tim and resolve them when they are resolved. I added some more. Also in PR message on its own you should only add to the section `Added` and do not describe stuff that changed within the PR.
Edit: If they are just regular comments and you cannot resolve them thumps up also works for me to see that you addressed them.

src/opi/input/structures/structure.py
```diff
@@ -1066,3 +1073,448 @@ def _iter_xyz_structures(
...
+        Parameters
+        ----------
+        elem_masses : dict[str | Element, float] | None, default: None
+            Per-element mass overrides keyed by element symbol string or
+            `Element` instance (e.g. `{"C": 13.003}`). If `None`, default
+            atomic masses are used.
+
+        Mass priority
+        -------------
+        elem_masses > atom.mass > default (ATOMIC_MASSES_FROM_ELEMENT)
```
haneug: Is this really reasonable? Should the atom mass not be prioritized?

haneug (approved):
Really nice PR! I just made some small style changes and will merge this now.

PR #236, opened 2026-04-29
feat: instance method to calculate MolBar from Structure

### LINKED ISSUE #227: Simple integration of MolBar into OPI´s Structure Object
I would like to interface OPI Structure objects with MolBar (https://git.rwth-aachen.de/bannwarthlab/molbar), a powerful molecular identifier.

A method `structure.to_molbar()` that delegates to a utility function in `opi/utils/molbar.py`. 
The utility module handles the lazy import of MolBar and hands the coordinates to Molbar with the `to_xyz_block()`. Secondly, we could add a standalone `molbar_data_to_structure()` utility function that reconstructs a `Structure` from a MolBar.

MolBar will not be added as a hard dependency. The utility module should raise a clear ImportError with an install hint if MolBar is not present, following the same pattern used for ASE.

We could add MolBar as a dependecy but I would not like to do that.

crjacinto:
## Closes Issues
Closes #227 

## Description
- Added an instance method that generates the Molecular Barcode from Structure.

### Added
- Added `calculate_molbar()` and `calculate_molbar_data()` to get Molbar from `Structure` (#236)


haneug:
@crjacinto please repair the title and the description of the PR

crjacinto:
> @crjacinto please repair the title and the description of the PR

Done

src/opi/input/structures/structure.py
```diff
@@ -861,6 +861,99 @@ def to_rdkitmol(self, structure: "Structure", /) -> "RdkitMol":
...
+        return_data: bool = False,
+        mode: str = "mb",
+        total_charge: int | None = None,
+    ) -> "str | tuple[str, dict[str, Any]]":
+        """
+        Compute the MolBar identifier for this :class:`Structure`.
+
+        (https://git.rwth-aachen.de/bannwarthlab/molbar).
+
+        Element symbols and Cartesian coordinates are extracted directly from
```
haneug: Please adhere to numpy docstring style. See other parts of the code for examples.

src/opi/input/structures/structure.py
```diff
@@ -861,6 +861,99 @@ def to_rdkitmol(self, structure: "Structure", /) -> "RdkitMol":
...
+        MolBar is **not** a dependency of OPI and must be installed separately::
+
+            pip install molbar
+
+        Parameters
+        ----------
+        return_data : bool, default False
+            If ``True``, return the full MolBar data dictionary in addition to
+            the barcode string.
+        mode : str, default ``"mb"``
```
haneug: Please set up a string enum to verify that valid string options are passed. Something like this:
```
class MolBarMode(StrEnum):
    MB = "mb"
    TOPO = "topo"
```

and then you can do:

```
  try:
      mode = MolBarMode(mode)
  except ValueError:
      raise ValueError(f"Invalid mode {mode!r}. Must be one of {[m.value for m in MolBarMode]}")
```

Either use `StrEnum` or use `StringEnum` from `opi/src/opi/models/string_enum.py` if it can be case insensitive.

src/opi/input/structures/structure.py
```diff
@@ -861,6 +861,99 @@ def to_rdkitmol(self, structure: "Structure", /) -> "RdkitMol":
...
+            MolBar calculation mode. ``"mb"`` computes the full barcode;
+            ``"topo"`` computes only the topology part.
+        total_charge : int | None, default None
+            Total charge passed to MolBar. When ``None``, defaults to
+            ``self.charge``.
+
+        Returns
+        -------
+        str
```
haneug: Please split this into two separate functions, one returning a `str`, one returning `tuple[str, dict[str, Any]`. A function that changes its return type based on a keyword argument makes type hints unreliable and is easy to misuse. Maybe make `to_molbar` return the string and `to_molbar_dict` or `to_molbar_full` return the whole data. 
Also please document or give the link to the documentation what this whole data even means.

src/opi/input/structures/structure.py
```diff
@@ -861,6 +861,99 @@ def to_rdkitmol(self, structure: "Structure", /) -> "RdkitMol":
...
+        except ImportError as err:
+            raise ImportError(
+                "MolBar is not installed. Install it with: pip install molbar"
+            ) from err
+
+        # > Collect elements and coordinates:
+        # > iterate self.atoms and call atom.format_xyz_line() on each entry.
+        # > Each line has the format "SYMBOL  x  y  z" (standard XYZ).
+        elements: list[str] = []
+        coords: list[list[float]] = []
```
haneug: Could this also be list[tuple[float,float,float]] since we only have x, y, z?

src/opi/input/structures/structure.py
```diff
@@ -861,6 +861,99 @@ def to_rdkitmol(self, structure: "Structure", /) -> "RdkitMol":
...
+
+        for atom in self.atoms:
+            parts = atom.format_xyz_line().split()
+            elements.append(parts[0])
+            coords.append([float(parts[1]), float(parts[2]), float(parts[3])])
+
+        if not elements:
```
haneug: You could check for `if self.atoms` before trying to loop over them and then the error message would fit better.

src/opi/input/structures/structure.py
```diff
@@ -861,6 +861,99 @@ def to_rdkitmol(self, structure: "Structure", /) -> "RdkitMol":
...
+            raise ValueError(
+                f"{self.__class__.__name__}.to_molbar: structure contains no atoms; "
+                "cannot build MolBar input."
+            )
+
+        # > Build (N, 3) coordinate array for type safety; convert back to list for MolBar
+        coordinates: list[list[float]] = np.array(coords, dtype=np.float64).tolist()
```
haneug: This step seems redundant to me. You already converted everything to floats?

src/opi/input/structures/structure.py
```diff
@@ -861,6 +861,99 @@ def to_rdkitmol(self, structure: "Structure", /) -> "RdkitMol":
...
+
+        Parameters
+        ----------
+        return_data : bool, default False
+            If ``True``, return the full MolBar data dictionary in addition to
+            the barcode string.
+        mode : str, default ``"mb"``
+            MolBar calculation mode. ``"mb"`` computes the full barcode;
+            ``"topo"`` computes only the topology part.
+        total_charge : int | None, default None
```
haneug: I would directly go with the default and I do not directly see the advantage of setting this via a keyword argument. One can always just change `self.charge`?

haneug (changes requested):
Nice PR! Please split the function according to the return types and change the docstring to numpy style. Also an enum for the mode would be a good thing for verification I think.

src/opi/input/structures/structure.py
```diff
@@ -861,6 +861,99 @@ def to_rdkitmol(self, structure: "Structure", /) -> "RdkitMol":
...
+
+        # > Collect elements and coordinates:
+        # > iterate self.atoms and call atom.format_xyz_line() on each entry.
+        # > Each line has the format "SYMBOL  x  y  z" (standard XYZ).
+        elements: list[str] = []
+        coords: list[list[float]] = []
+
+        for atom in self.atoms:
+            parts = atom.format_xyz_line().split()
+            elements.append(parts[0])
```
haneug: This could be problematic when you have fragments in your structure. Please check that as well and use `atom.element` instead!

haneug: Also you should make sure to just include actual atoms and not e.g., point charges do that by checking if it is an instance of `Atom`

haneug: and use `atom.coordinates` for x,y,z

src/opi/input/structures/structure.py
```diff
@@ -861,6 +861,99 @@ def to_rdkitmol(self, structure: "Structure", /) -> "RdkitMol":
...
+
+        Raises
+        ------
+        ImportError
+            If MolBar is not installed.
+        ValueError
+            If this structure contains no real :class:`~opi.input.structures.atom.Atom`
+            instances.
+        """
+        try:
```
timmyte: Good PR. Thanks for adding this MolBar.

Import statements should always be on module level. There are more less only two reasons to do them on a lower level:
1) For performance, e.g. to speed up CLI auto-completion.
2) If it would lead to a circular import.

It does not seem to me, that either criteria is met here.

The import could look as follows:
```
from functools import wraps

try:
  from molbar.barcode import get_molbar_from_coordinates
except ImportError:
  # > Making sure the identifier is the global namespace
  get_molbar_from_coordinates = None

def requires_molbar(func):
    @wraps(func)
    def wrapper(*args, **kwargs):
      if get_molbar_from_coordinates is None:
        raise raise ImportError("MolBar is not installed. Install it with: pip install molbar")
      return func(*args, **kwargs)
    return wrapper

# [...]

@requires_molbar
def to_molbar(...):
```

src/opi/input/structures/structure.py
timmyte: How about adding a `Structure.get_coordinates()` which returns all coordinates as single numpy array.
This can easily be converted to a list of lists using `ndarray.tolist()`.

src/opi/input/structures/structure.py
crjacinto: Done

src/opi/input/structures/structure.py
crjacinto: Behavior changed in new implementation. Thanks for the correction

src/opi/input/structures/structure.py
crjacinto: Yes, changed

src/opi/input/structures/structure.py
crjacinto: The new functions `get_coordinates()` takes care of this now. Thanks for the observation

src/opi/input/structures/structure.py
crjacinto: Done, thanks for the explanation

crjacinto:
At this point, the latest commit of this code is compatible with the features developed for the RMSD-RotConst implementation (#230). The unit tests are failing since some of the features that are being used now come from the RMSD-RotConst code which has not yet been merged here. Once the other code is merged, I'm going to rebase this commit and it would be ready for review

src/opi/input/structures/structure.py
```diff
@@ -1110,6 +1111,123 @@ def from_lists(
...
+        See the MolBar documentation for a full description of the barcode format.
+
+        Only real `Atom` entries are passed to MolBar; `EmbeddingPotential`,
+        `GhostAtom`, and `PointCharge` instances are silently skipped. Note
+        that `GhostAtom` is a subclass of `Atom`, so an exact type check is
+        used rather than `isinstance`.
+
+        The total charge is taken from `charge`.
+
+        MolBar is **not** a dependency of OPI and must be installed separately::
```
haneug: Here you write molbar is not a dependency and below you write it is optional. We should clearly decide for one thing and go with it. Please make it an optional dependency and pin a version number we want to use.

src/opi/input/structures/structure.py
```diff
@@ -1110,6 +1111,123 @@ def from_lists(
...
+    # ------------------------------------------------------------------ #
+
+    @requires_molbar
+    def calculate_molbar(
+        self,
+        *,
+        mode: "MolBarMode | str" = MolBarMode.MB,
+    ) -> str:
+        """
+        Compute the MolBar barcode string for this `Structure`.
```
haneug, line 1125: Could we add the link to the paper here? With the doi?

src/opi/utils/molbar.py
```diff
@@ -0,0 +1,126 @@
...
+)
+
+# ============================================================
+# MolBar optional import
+# ============================================================
+
+# MolBar is an optional dependency. The decorator below guards every method
+# that calls it and raises a clear ImportError if it is not installed.
+try:
+    from molbar.barcode import get_molbar_from_coordinates  # type: ignore
```
haneug: This makes molbar get imported whenever `opi.utils.molbar` is imported and since structure.py imports it, that means on any `from opi.core import Calculator` etc. it will be imported. Molbar is a large package with heavy dependencies, so for users who have it installed this slows down every OPI import even if they never compute a barcode. Suggestion: move the import into `call_molbar()` so it only happens on first use. Note that `requires_molbar` currently relies on this module-level import (it checks `get_molbar_from_coordinates()` is None), so it would need to switch to something like `importlib.util.find_spec("molbar")`.

tests/unit/test_structure_molbar.py
```diff
@@ -0,0 +1,341 @@
...
+
+    def test_raises_for_invalid_mode(self, water):
+        """`calculate_molbar_data()` should raise `ValueError` for an invalid mode."""
+        with pytest.raises(ValueError, match="Invalid mode"):
+            water.calculate_molbar_data(mode="invalid")
+
+    def test_topo_mode(self, water):
+        """`calculate_molbar_data()` with `mode="topo"` should return the topology barcode."""
+        barcode, _ = water.calculate_molbar_data(mode="topo")
+        assert barcode == "TopoBar | 1.1.3 | OH2 | -84 20 344 | 80"
```
haneug: can we replace `1.1.3` by `molbar.__version__`? Is that possible?

src/opi/input/structures/structure.py
haneug: I added it as an optional dependency please document it accordingly.

haneug (changes requested):
Nice work, I would suggest to move the molbar import from the module level to the function level. Also, please add an entry to `CHANGELOG.md` and I think there was one comment about `to_molbar` which is not anymore.

tests/unit/test_structure_molbar.py
haneug: When I looked at `molbar.__version__)` it raised an `AttributeError`

src/opi/input/structures/structure.py
crjacinto: Done, thanks

src/opi/utils/molbar.py
crjacinto: Changed the behavior of the import, thanks for the correction

src/opi/input/structures/structure.py
crjacinto: Done

tests/unit/test_structure_molbar.py
crjacinto: Changed. Now we read the version at module level using metadata. I also made it more flexible so we can start testing the newest version. I also changed the hard coded parts so they take the string obtained from the metadata

src/opi/input/structures/structure.py
```diff
@@ -1078,7 +1079,7 @@ def from_lists(
 
         Parameters
         ----------
-        symbols : list[str | int]
+        symbols : listMOLBAR_VERSION = version("molbar") at module level, using importlib.metadata.version instead of molbar.__version__. This reads from installed package metadata, so it works even though MolBar doesn't expose __version__[str | int]
```
haneug: Here the docstring seems to be broken!

src/opi/input/structures/structure.py
```diff
@@ -1110,6 +1111,125 @@ def from_lists(
...
+    @requires_molbar
+    def calculate_molbar(
+        self,
+        *,
+        mode: "MolBarMode | str" = MolBarMode.MB,
+    ) -> str:
+        """
+        Compute the MolBar barcode string for this `Structure`.
+
+        <https://doi.org/10.1039/d4dd00208c>
```
haneug: please give a little bit more context to what this is, e.g. Original Molbar publication: ...

haneug (changes requested):

src/opi/input/structures/structure.py
crjacinto: Sorry, some weird typo from side. Fixed now

src/opi/input/structures/structure.py
crjacinto: Done, thanks for the observation

src/opi/input/structures/structure.py
```diff
@@ -1110,6 +1111,132 @@ def from_lists(
...
+        MolBar (Molecular Barcode) was introduced by van Staalduinen and
+        Bannwarth as a chemical identifier that overcomes limitations of
+        SMILES and InChI for inorganic molecules and non-central stereochemistry
+        (e.g. axial and planar chirality). It combines the conventional
+        atomistic description with a fragment-based approach: fragment 3D
+        structures are normalised with a specialised force field and
+        characterised by physically inspired matrices derived solely from
+        atomic positions. The resulting permutation-invariant representation
+        is built from the eigenvalue spectra of these matrices, encoding both
+        bonding and stereochemistry. See the original publication for details:
```
timmyte, line 1138: Please add some line breaks for better readability. Otherwise very nicely documented.

```suggestion
        MolBar (Molecular Barcode) was introduced by van Staalduinen and Bannwarth as a
chemical identifier that overcomes limitations of SMILES and InChI for inorganic
molecules and non-central stereochemistry (e.g. axial and planar chirality).
It combines the conventional atomistic description with a fragment-based approach:
fragment 3D structures are normalised with a specialised force field and
characterised by physically inspired matrices derived solely from atomic
positions.
The resulting permutation-invariant representation is built from the
eigenvalue spectra of these matrices, encoding both bonding and stereochemistry.
See the original publication for details:
```

src/opi/input/structures/structure.py
```diff
@@ -1424,6 +1551,84 @@ def _iter_xyz_structures(
...
+        if not self.real_atoms:
+            raise ValueError(
+                f"{self.__class__.__name__}: structure contains no real atoms; "
+                "cannot build MolBar input."
+            )
+
+        # Only real Atom instances are passed to MolBar; EmbeddingPotential,
+        # GhostAtom, and PointCharge entries are excluded via real_atoms.
+        # Note that GhostAtom is a subclass of Atom, so type(a) is Atom is
+        # used rather than isinstance inside the real_atoms property.
```
```suggestion
        # > Only real Atom instances are passed to MolBar; EmbeddingPotential,
        # > GhostAtom, and PointCharge entries are excluded via real_atoms.
        # > Note that GhostAtom is a subclass of Atom, so type(a) is Atom is
        # > used rather than isinstance inside the real_atoms property.
```
To separate comment out code from actual comments ;)

src/opi/input/structures/structure.py
```diff
@@ -1424,6 +1551,84 @@ def _iter_xyz_structures(
...
+        -------
+        MolBarMode
+            The validated and normalised mode.
+
+        Raises
+        ------
+        ValueError
+            If *mode* is not a valid `MolBarMode` value.
+        """
+        # Normalise and validate mode (case-insensitive via StringEnum._missing_)
```
```suggestion
        # > Normalise and validate mode (case-insensitive via StringEnum._missing_)
```

src/opi/utils/molbar.py
```diff
@@ -0,0 +1,134 @@
...
+__all__ = (
+    "MolBarMode",
+    "requires_molbar",
+)
+
+# ============================================================
+# MolBar optional import
+# ============================================================
+
+# > MolBar is an optional dependency with heavy dependencies. It is imported
```
timmyte: Very good idea!!

src/opi/utils/molbar.py
```diff
@@ -0,0 +1,134 @@
...
+# ============================================================
+# Mode classification
+# ============================================================
+
+
+class MolBarMode(StringEnum):
+    """Valid calculation modes for `Structure.calculate_molbar`.
+
+    Matching is case-insensitive: pass `"MB"`, `"mb"`, or `"Mb"`
+    and all will be accepted.
```
timmyte: I would rather refer to Molbar's documentation of the calculation mode.
```suggestion
    """
    Molbar's calculations modes.
    More details about these can be found: ...
    
    Entries in the enum can be resolved case-insensitively.
```

src/opi/utils/molbar.py
```diff
@@ -0,0 +1,134 @@
...
+
+    Attributes
+    ----------
+    MB : str
+        Full MolBar barcode (`"mb"`).
```
timmyte:
```suggestion
        Full MolBar barcode (`mb`).
```

src/opi/utils/molbar.py
```diff
@@ -0,0 +1,134 @@
...
+    TOPO : str
+        Topology-only barcode (`"topo"`).
```
timmyte:
```suggestion
        Topology-only barcode (`topo`).
```

src/opi/utils/molbar.py
```diff
@@ -0,0 +1,134 @@
...
+# MolBar optional import
+# ============================================================
+
+# > MolBar is an optional dependency with heavy dependencies. It is imported
+# > lazily inside `call_molbar()` so that merely importing `opi.utils.molbar`
+# > (which happens transitively on almost any OPI import via structure.py)
+# > does not pull in MolBar for users who never compute a barcode.
+
+
+def _molbar_available() -> bool:
```
timmyte: I would just move this function next to the decorator, as this routine is basically a part of the decorator.

src/opi/utils/molbar.py
```diff
@@ -0,0 +1,134 @@
...
+    Returns
+    -------
+    str | tuple[str, dict[str, Any]]
+        Either the barcode string or the barcode string together with the
+        full MolBar data dictionary, depending on *return_data*.
+    """
+
+    # > Imported lazily: this is the first point at which MolBar is actually
+    # > needed, so users who never compute a barcode never pay the import cost.
+    from molbar.barcode import get_molbar_from_coordinates  # type: ignore
```
timmyte: Almost perfect!!
If you import the module in the function, it will be imported every time the function is called and discarded afterwards.
So performance-wise this even worse the importing it once globally and never using it.
I would defer the import into a decorator for this function here using `import_module()` and add the module to the `globals()`.
Thereby the module is imported once when first needed and kept in memory afterwards.

Just make sure to mention that this is not thread-safe.

pyproject.toml
```diff
@@ -37,6 +37,11 @@ Repository = "https://github.com/faccts/opi"
 Issues = "https://github.com/faccts/opi/issues"
 Changelog = "https://github.com/faccts/opi/blob/main/CHANGELOG.md"
 
+[project.optional-dependencies]
+molbar = [
+    "molbar==1.1.3",
```
timmyte: @haneug 
Usually I wouldn't bolt down the version restriction to a very specific version, but just allow for bugfix release, so in that case `<=1.1`

src/opi/utils/molbar.py
```diff
@@ -0,0 +1,134 @@
...
+
+    return wrapper
+
+
+# ============================================================
+# Free helper function
+# ============================================================
+
+
+def call_molbar(
```
timmyte, line 128: Very nice interface!
But I would move the conversion from OPI specific datatype to common ones understood by Molbar to here, i.e.
1) `elements` could also be a list of `Element`
2) `coordinates` could also be a numpy array.

I would always do these type conversion as late as possible.

timmyte (changes requested):
Very nice addition. It's great how well the code is documented at this point!!

src/opi/input/structures/structure.py
crjacinto: Good, thanks

src/opi/utils/molbar.py
crjacinto: Danke

src/opi/input/structures/structure.py
crjacinto: Done, thanks

src/opi/utils/molbar.py
crjacinto: Changed, thanks

src/opi/utils/molbar.py
crjacinto: Done, thanks

src/opi/utils/molbar.py
crjacinto: Added link, thanks for the suggestion

src/opi/utils/molbar.py
crjacinto: Changed, thanks

src/opi/utils/molbar.py
crjacinto: Just created a dedicated `@_import_molbar` decorator on `call_molbar`, that is kept separated from the existing `requires_molbar` guard.

On the first call to `call_molbar`, the decorator does `import_module("molbar.barcode")` and adds `get_molbar_from_coordinates` to `globals()`. In this way, future calls skips the import and the heavy dependencies are imported only when a first barcode is computed.

`get_molbar_from_coordinates` is declared at module level as `Any = None` just for the type checkers . On the first use, the decorator flips it from None to the imported function. 

The two decorators stay separate by "function": `requires_molbar` (on the Structure methods) performs the availability check and raises a clean `ImportError` if MolBar is missing, without importing anything; `_import_molbar` (on `call_molbar`) performs the actual cached import. 

As suggested, the docstring notes that the first call to the import-and-cache step is not thread-safe: concurrent first calls could run the import more than once, which is harmless (the module would be already cached in `sys.modules` and the resolved function is identical) but is documented explicitly.

pyproject.toml
crjacinto: Since version 1.2 is around the corner, I defined it as ~=1.1 so it allows bug fixes and backward-compatible updates

src/opi/utils/molbar.py
crjacinto: I passed the conversions to the Molbar module. Now `structure.py` handles only native OPI data. Thanks for the observation.

src/opi/utils/molbar.py
```diff
@@ -0,0 +1,178 @@
...
+
+
+# ============================================================
+# Free helper function
+# ============================================================
+
+
+@_import_molbar
+def call_molbar(
+    elements: list[Element] | list[str],
```
timmyte:
```suggestion
    elements: Sequence[Element, str]
```

src/opi/utils/molbar.py
```diff
@@ -0,0 +1,178 @@
...
+
+    Returns
+    -------
+    str | tuple[str, dict[str, Any]]
+        Either the barcode string or the barcode string together with the
+        full MolBar data dictionary, depending on *return_data*.
+    """
+
+    # > Convert OPI-native types to the plain types MolBar understands.
+    element_symbols: list[str] = [e.value if isinstance(e, Element) else e for e in elements]
```
timmyte:
```suggestion
    element_symbols: list[str] = [Element(e) for e in elements]
```
should also work and is a bit smoother.

src/opi/utils/molbar.py
```diff
@@ -0,0 +1,178 @@
...
+    coords: list[list[float]] = np.asarray(coordinates, dtype=np.float64).tolist()
```
timmyte: I would casting a list blindly to array. This is relatively expensive.
I would go for:

```suggestion
    coords: list[list[float]] = coordinates.tolist() if isinstance(coordinates, numpy.ndarray) else coordinates
```

src/opi/utils/molbar.py
```diff
@@ -0,0 +1,178 @@
...
+
+__all__ = (
+    "MolBarMode",
+    "requires_molbar",
+)
+
+# > Populated on first use by the `@_import_molbar` decorator (see below).
+# > Declared here so the name resolves for type checkers and at call time,
+# > before the first invocation has cached the real function.
+get_molbar_from_coordinates: Any = None
```
timmyte: I would be a bit more precise on the the type annotation:
```
# > For exact function signature, consult Molbar's documentation.
get_molbar_from_coordinates: Callable[[list[list[float]], list[str | int]], str | tuple] | None = None
```

timmyte (approved):
Awesome. Looks good know. Just some last minor improvements.

src/opi/utils/molbar.py
crjacinto: Changed, thanks

src/opi/utils/molbar.py
crjacinto: As the suggestion stands now, the return type would be `list[Element]`, which doesn't agree with the annotation. Also, Molbar expects a list of strings so this could misbehave. 

Based on your suggestion, I wrote instead:

`element_symbols: list[str] = [Element(e).value for e in elements]`

This normalizes both `str` and `Element` so we get the correct input for Molbar.

src/opi/utils/molbar.py
crjacinto: Thanks for the comment. I was checking this, and in fact, Molbar expects only an array with the coordinates that has dimensions `natoms` x 3. So, the nested lists are not that necessary.

I have double checked, and Molbar works with both data structures: nested lists or arrays, that is why I changed this line directly to:

`coords = np.asarray(coordinates, dtype=np.float64)` 

which should be the cheapest option to handle different data type inputs.

src/opi/utils/molbar.py
crjacinto: Thanks for the comment. Since the return types have changed, and also due to the fact that for a `Callable` can't handle keyword arguments (we use total_charge, return_data, mode), I rewrote this line using unconstrained args.

`get_molbar_from_coordinates: Callable[..., str | tuple[str, dict[str, Any]]] | None = None`

Also consider that this line exists to resolve the type checker and the decorator afterwards, so going beyond this (like a `Protocol`) is unnecessary in my opinion.

crjacinto: I also added an `assert` statement since mypy was flagging the Callable | None discrepancy

haneug (approved):

PR #238, opened 2026-05-01
Updated `BaseRunner.run` API

### LINKED ISSUE #235: Simplify `BaseRunner.run` method
I am working through issue #224 and I think the current `BaseRunner.run` output API could be simplified.

My understanding of the current behavior is:

- `capture=True` uses `subprocess.PIPE` so `stdout` and `stderr` are returned on the `CompletedProcess`.
- `silent=True` redirects output to `os.devnull`, unless overridden.
- `stdout` / `stderr` can be set to `Path` objects to write the corresponding stream to disk.

This creates a few ambiguous combinations:

- `capture=True, silent=True` silently ignores `silent`.
- `capture=False, stdout=path` or `stderr=path` still redirects that stream to a file, despite capture being disabled.
- `stdout` / `stderr` are currently path-oriented, so users cannot pass an open file object, in-memory buffer, callback, logger, or multiple destinations.
- Internal code that needs to inspect stderr must either force `capture=True` or write stderr to a temporary file and read it back.

This is particularly awkward for `OrcaMmRunner.run_orca_mm`, which needs to inspect stderr to decide whether to raise `OrcaMmException`. At the moment, that requires coupling the internal error-checking behavior to the public output-routing behavior.

I think the issue is that `capture`, `silent`, `stdout`, and `stderr` are all competing controls for the same underlying concern: stream handling.

## Proposal

Replace the current flag-based API with a stream-target API.

The runner would always capture `stdout` and `stderr` internally and return them on the result. The `stdout` and `stderr` arguments would only describe additional destinations for each stream.

For example:

```python
StreamTarget = (
    None
    | Path
    | IO[str]
    | Callable[[str], None]
    | Sequence["StreamTarget"]
)
```

Proposed semantics:
- `None`: no external output destination
- `Path`: write the stream to a file
- `IO[str]`: write the stream to an open file-like object
- `Callable[[str], None]`: stream chunks to a callback
- `Sequence[StreamTarget]`: tee the stream to multiple destinations

The run method could then look something like:
```python
def run(
    self,
    binary: OrcaBinary,
    args: Sequence[str] = (),
    *,
    stdout: StreamTarget = None,
    stderr: StreamTarget = None,
    stdin: str | None = None,
    cwd: Path | None = None,
    timeout: int | None = None,
) -> RunResult:
    ...
```

Where `RunResult` exposes the captured output:
```python
@dataclass
class RunResult:
    binary: str
    args: Sequence[str]
    returncode: int
    stdout: str
    stderr: str
```

This removes the need for both `capture` and `silent`:
```python
# Current silent behavior
runner.run(..., stdout=None, stderr=None)

# Write stderr to a file while still allowing internal inspection
result = runner.run(..., stderr=Path("orca_mm.err"))

# Print stdout live
result = runner.run(..., stdout=lambda chunk: print(chunk, end=""))

# Print stdout live and write it to disk
result = runner.run(
    ...,
    stdout=[
        lambda chunk: print(chunk, end=""),
        Path("orca.out"),
    ],
)
```

For `OrcaMmRunner.run_orca_mm`, this would simplify the implementation because stderr would always be available on the returned result:
```python
result = self.run(
    OrcaBinary.ORCA_MM,
    args,
    stdout=stdout,
    stderr=stderr,
)

if raise_on_error and result.stderr:
    raise OrcaMmException(command, arguments, result.stderr)
```

This would decouple internal validation from user-facing output routing. Users can decide where output should go, while the runner can still inspect process output consistently.

One possible caveat is memory usage for very large outputs, since this proposal implies always retaining `stdout` and `stderr` in memory. If that becomes a concern, we could add a later escape hatch such as `max_capture_bytes`, but I think the simpler default API is preferable unless large-output cases become a real problem.

MattBurn:
## Closes Issues
Closes #235

## Description
- Replaces the competing `capture`, `silent`, `stdout`, and `stderr` controls in `BaseRunner.run` with explicit `stdout`/`stderr` stream targets.
- Adds subprocess stream fanout support for `Path`, string paths, open text streams, callbacks, `subprocess.PIPE` capture, and multiple destinations using `opi.execution.text_stream.TextStreamFanout`.
- Adds immutable run result objects `opi.execution.base.RunResult` and `opi.execution.run.SubprocessRunResult` with captured output, return-code helpers.
- Deprecates `stdin_str`, `capture`, and `silent` in favor of `stdin` and `stdout`/`stderr` stream targets.
- Adds focused unit tests for capture, fanout, live callbacks, stdin, timeout handling, and invalid stream targets.

---

### Added

- Added `TextStreamFanout` stream-target fanout support for subprocess stdout and stderr destinations. (#238)
- Added immutable subprocess run result objects with return-code and signal helpers. (#238)

### Changed

- Changed `BaseRunner.run` to use explicit stdout and stderr stream targets. (#238)
- Changed `BaseRunner.run` to return a `RunResult` instead of `subprocess.CompletedProcess` or `None`. (#238)

### Deprecated

- Deprecated `stdin_str`, `capture`, and `silent` on `BaseRunner.run`. (#238)


MattBurn:
Apologies for the number of CI runs, finally figured out how nox works to check locally in the future

timmyte:
> Apologies for the number of CI runs, finally figured out how nox works to check locally in the future

Anything that could be improved in the docs to make this clearer?

MattBurn:
I am new to nox so I did not grasp what nox was going to run in CI before pushing. I ended up using `uv tool run nox` which was a nice and simple way to run everything at once locally. Possibly worth adding to the development guide.

MattBurn:
Are there any blockers for getting this merged?

src/opi/execution/base.py
```diff
@@ -73,6 +75,40 @@ def wrapper(self: RunnerType, /, *args: Any, **kwargs: Any) -> R:
...
+    def returncode_ok(self) -> bool:
+        """Check for zero exit code"""
+        return self.returncode == 0
+
+    def check_returncode(self) -> None:
+        """Raise RuntimeError exit code is non-zero."""
+        if not self.returncode_ok():
+            raise RuntimeError(
+                f"Running '{self.binary}' with arguments {self.args} failed with exit code: {self.returncode}"
+            )  # change to OpiExecutionError when PR #224 merged
```
timmyte: Please adjust the comment as follows
```suggestion
            )  # > TODO: change to OpiExecutionError when PR #224 merged
```

src/opi/execution/base.py
```diff
@@ -261,10 +302,15 @@ def get_version(self) -> OrcaVersion | None:
...
+            orca_proc = self.run(
+                OrcaBinary.ORCA,
+                ["--version"],
+                stdout=subprocess.PIPE,
+                timeout=5,
+            )
 
             # > Pleasing type checker
-            assert isinstance(orca_proc, CompletedProcess)
+            # assert isinstance(orca_proc, CompletedProcess)
```
timmyte: Please remove the assert statement completely.
```suggestion
```

timmyte (approved):
@MattBurn 

Very sorry for the long waiting time. I was quite busy.

Looks good. Could you please make the little corrections I remarked before merging. Thanks!!

timmyte:
@haneug 
As I made tiny changes, would you quickly going over this as well

@timmyte 
- [ ] Run testsuite before merge.

haneug (approved):

timmyte:
@haneug 
Please review again, as made significant changes and additions.

@MattBurn 
If you like to also review, by my guest. I would love to hear your opinion.
I'm very sorry again, for the long delay. It's very nicely structured and documented code.
I enjoyed working with it.

MattBurn:
Changes look good to me, I originally started this to improve the ergonomics of creating the orca_mm binary in PR #224. With so many changes to how the runner works, do you think it would be best to restart this PR?

timmyte:
> Changes look good to me, I originally started this to improve the ergonomics of creating the orca_mm binary in PR #224. With so many changes to how the runner works, do you think it would be best to restart this PR?

Thanks for your feedback.
From our side there's no need to restart the PR — but feel free to if you'd prefer. I can imagine merging the changes into #224 would be a lot of work, so under the circumstances a rebase and force-push is fine too.

haneug (approved):

timmyte:
Had to make some changes. Units test pass fine now.
@MattBurn thanks for adding them so rigorously.

PR #241, opened 2026-05-20
Add RHF and ROHF to get_s2 getter function

### LINKED ISSUE #239: Also get a return value for S2 for simple models
**description**
The function `get_s2` only returns a value for unrestricted calculations and otherwise None. I would prefer it also to return the value for $S^2$ in case of restricted calculations and not just None.

**possible solution**
I suppose one could just look at the `Hftype` from `get_hftype()`:
- if Hftype.RHF -> 0
- if Hftype.ROHF -> `(get_mult()**2 - 1)/4`

haneug:
## Closes Issues
Closes #239 

## Description
- Add S^2 return value for the simple models RHF (just 0.0) and ROHF (ideal value from the multiplicity) to the getter function `get_s2`

---

### Added 

- The function output `get_s2` now returns values for RHF/RKS and ROHF/ROKS (#241).

### DIFF
--- a/src/opi/output/core.py
+++ b/src/opi/output/core.py
@@ -1779,26 +1779,47 @@ class Output:
 
     def get_s2(self, *, index: int = -1) -> tuple[StrictFiniteFloat, StrictFiniteFloat] | None:
         """
-        Get the S² expectation value and the ideal S² value by grepping them from the output file.
+        Get the S² expectation value and the ideal S² value by grepping them from the output file for UHF and returning
+        the ideal values for RHF and ROHF.
 
         Parameters
         ----------
         index : int, default: -1
             Index of the geometry for which S² should be returned. The default -1 refers to the final geometry.
             Silently ignores if the requested index is not available and returns None.
+            For RHF/ROHF the index is not relevant since the ideal values are returned.
 
         Returns
         ----------
         tuple[StrictFiniteFloat, StrictFiniteFloat] | None
             Return the expectation value and the ideal value or None if nothing is found.
         """
-        outfile = self.get_outfile()
-        # // String for searching the S² expectation value.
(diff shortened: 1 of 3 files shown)


src/opi/output/core.py
```diff
@@ -1837,26 +1837,47 @@ def get_polarizability(self, *, index: int = -1) -> list[Polarizability] | None:
...
+        # > Get HFType
+        hftype = self.get_hftype()
+        match hftype:
+            # > RHF/RKS just returns the ideal value zero
+            case Hftyp.RHF:
+                return 0.0, 0.0
+            # > ROHF/ROKS - ideal S2 value by construction
+            case Hftyp.ROHF:
+                mult = self.get_mult()
+                # > Cannot determine S2 without multiplicity
```
timmyte, line 1869: Please mention this also in the first part of the docstring.

src/opi/output/core.py
```diff
@@ -1837,26 +1837,47 @@ def get_polarizability(self, *, index: int = -1) -> list[Polarizability] | None:
 
     def get_s2(self, *, index: int = -1) -> tuple[StrictFiniteFloat, StrictFiniteFloat] | None:
```
timmyte: I wouldn't return just `None` in the case nothing is found, as in the case something was found, it returns a well defined number of values. So users might be tempted to unpack them directly:
```
a, b = get_s2()
```
which will raise a `TypeError` if just `None` is returned.
Instead I would change the return behavior to:
```
->  tuple[StrictFiniteFloat, StrictFiniteFloat] | tuple[None, None]
```

timmyte (approved):
This is a nice, small addition. Thanks.
Please regard my two small remarks before merging.

PR #243, opened 2026-05-22
WIP: Additional integral getters

### LINKED ISSUE #240: Adding getters for other electronic integrals
OPI currently has getters for some of the integrals available from ORCA such as `get_int_hcore`, `get_int_overlap` however it lacks similar getters for some of the other integrals, such as `HMO`, `T` and `V` as well as two electron integrals, which presumably could also be exposed to the user similarly.

Additional getters in `src/opi/output/core.py` for other 1 electron integrals and 2 electron integrals, in a similar format to the existing getters.

As an alternative it is possible to directly extract this information from the GBW JSON data, however this negates the nice and safe getters introduced in OPI.

**Additional context**
I'm not sure if there is a particular reason why these other integrals might not be implemented yet, but if it is simply a matter of not having gotten round to it quite yet, I would also be happy to make a PR following in the form of the other getters, which you can then tweak as you find suitable.

Thanks for your time!

vaferreiQMT:
## Closes Issues
Closes #240

## Description
- Adds `t_matrix`, `v_matrix` and `hmo` to `Molecule` along with corresponding getters in `src/opi/output/core.py` (`get_int_kinetic`, `get_int_nuclear`, `get_int_hmo`)
- Adds `TwoElectronIntegralElement` type for the sparse COO 4 index matrix elements
- Adds `MolecularTwoElectronIntegral` class for two electron integrals in MO basis
- Adds `TwoElectronIntegrals` class for all two electron integral information and corresponding `twoelintegrals` in `Molecule` class

I have attempted to reproduce the style and format of the existing code but of course a lot of it is also up to taste in terms of how things are arranged and named so very happy for you to change it as you see fit.

---

### Added 
- Added `t_matrix`, `v_matrix`, `hmo`, `twoelintegrals` parameters to `Molecule` class (#243)
- Added `get_int_kinetic`, `get_int_nuclear`, `get_int_hmo` getters for one electron integrals (#243)
- Added `TwoElectronIntegralElement`, `MolecularTwoElectronIntegral` and `TwoElectronIntegrals` for structuring two electron integral information (#243)


vaferreiQMT:
Although the additional objects added here are sufficient to extract and parse the two electron integral information (at least from my testing), I have not yet gone ahead and added getters. Given there are so many possible two electron integrals I'm not sure if you wish to have separate getters as is currently done or go for a different approach, but I wanted to check before adding hundreds of lines of boilerplate.

haneug:
Thank you very much for your nice PR. The unit tests `uv run nox -s unit_tests` are currently failing because the new pydantic fields are not covered by our JSON file fixtures. 
You can either add the new fields to the list of attributes to ignore which can be found here: 
https://github.com/faccts/opi/blob/821671e4264993c6bbee591b74d8fd342cdfbdb5/tests/unit/test_output_attributes.py#L46
Or, if you would provide an example that makes use of the new pydantic fields and getters we could add it to the JSON fixtures and cover them in the attribute output tests.

vaferreiQMT:
Thanks for the help! I've updated the integrals test example with the additional single electron getters and added the two electron integrals to the ignore list for now. Unit tests now seem to pass locally.

src/opi/output/core.py
```diff
@@ -2159,6 +2159,111 @@ def get_int_hcore(
         else:
             return None
 
+    def get_int_kinetic(
+        self, recreate_json: bool = False, gbw_index: int = 0
+    ) -> npt.NDArray[np.float64] | None:
+        """
+        Returns the kenetic energy integral matrix as numpy array.
```
haneug: Small typo: kenetic.

src/opi/output/core.py
```diff
@@ -2159,6 +2159,111 @@ def get_int_hcore(
...
+
+        # > get kinetic integrals from gbw json files
+        kinetic_list = self._safe_get("results_gbw", gbw_index, "molecule", "t_matrix")
+
+        if kinetic_list is not None:
+            return np.array(kinetic_list)
+        else:
+            return None
+
+    def get_int_nuclear(
```
haneug: I think the name `get_int_nuclear` is a bit ambiguous although it should be clear what is meant. To be a bit clearer, I would suggest to use `get_int_nuc_attr`. But this would just be a suggestion.

src/opi/output/models/json/gbw/properties/molecular_two_electron_integral.py
```diff
@@ -0,0 +1,28 @@
...
+
+from opi.output.models.base.get_item import GetItem
+from opi.output.models.json.gbw.properties.two_electron_integral_element import (
+    TwoElectronIntegralElement,
+)
+
+
+class MolecularTwoElectronIntegral(GetItem):
+    """
+    This class contains the information about two electron integrals in MO basis
```
haneug, line 11: The naming is here a little bit off I think since the class is named `MolecularTwoElectronIntegral` and it holds two electron integrals in the molecular orbitals basis. Maybe go with `MOTwoElectronIntegral` instead?

haneug (changes requested):
Very Nice PR, I only found a small typo and have two suggestions for slightly different naming. I did not come to thoroughly check the two electron integrals but from what I see it looks good.

src/opi/output/core.py
vaferreiQMT: Corrected in [2f22fdf](https://github.com/faccts/opi/pull/243/commits/2f22fdfdc0cb0fbfc5c6ad2a95eabe133aa752c7).

src/opi/output/core.py
vaferreiQMT: Sounds good to me, updated in [1a18172](https://github.com/faccts/opi/pull/243/commits/1a18172418e537a00daff6034d03c948746f8f44).

src/opi/output/models/json/gbw/properties/molecular_two_electron_integral.py
vaferreiQMT: Updated in [0e20671](https://github.com/faccts/opi/pull/243/commits/0e20671626b3b6be5a2a2f7734e46fba70014232).

haneug (approved):

haneug:
Nice, I would merge your PR. If you could also add your release note section to the `CHANGELOG.md` file before that. That would be great.

vaferreiQMT:
Great, updated the changelog now! Just wanted check as well if you had any thoughts on the format for the two electron integral getters (which currently aren't included here).

haneug:
@vaferreiQMT Thanks for the additions to the changelog. There seems to be an issue with the license. That sometimes happens and in my experience an empty commit can resolve the issue sometimes.

haneug:
Thank you for contributing to OPI and making it better. Regarding the two electron integral getters I do not have a strong opinion and would just expose them as ORCA supplies them. If you require anything specific / any specific convention we would love to hear it!

PR #247, opened 2026-06-01
feat: implement frequency getters

### LINKED ISSUE #246: Add getter function for frequencies and imaginary frequencies.
Add a getter function which returns all frequencies after a frequency calculation `get_frequencies` in a dict of type `dict[int, float]` and add a subsequent getter function that only returns the dict of imaginary frequencies `get_imaginary_frequencies`

haneug:
## Closes Issues
Closes #246

## Description
- Implements new getter functions that work with frequency calculations.
- Includes `get_frequencies`, `get_imaginary_frequencies`, `is_pes_minimum`, and `is_pes_transition_state`.
- Comes with a new example 56 that demonstrates how to use these in a transition state optimization.

---

### Added 

- Added `get_frequencies`, `get_imaginary_frequencies`, `is_pes_minimum`, and `is_pes_transition_state` to the `Output` class (#247).


src/opi/output/core.py
```diff
@@ -2198,6 +2198,96 @@ def get_free_energy_delta(self, *, index: int = -1) -> StrictFiniteFloat | None:
...
+            data is present, and an empty dict if all frequencies are real.
+        """
+        frequencies = self.get_frequencies()
+
+        if frequencies is None:
+            return None
+
+        return {mode: freq for mode, freq in frequencies.items() if freq < 0}
+
+    def is_pes_minimum(self, *, threshold: float = 0.0) -> bool | None:
```
timmyte, line 2257: I would also add the `threshold` to `get_frequencies()` and `get_imaginary_frequencies()` with `None` as default.
If no modes can be found at all I would still return `None`, but return `{}` if all modes were filtered out, due to the threshold.

src/opi/output/core.py
```diff
@@ -2198,6 +2198,96 @@ def get_free_energy_delta(self, *, index: int = -1) -> StrictFiniteFloat | None:
...
+        is_minimum : bool | None
+            True if no frequency lies below ``-threshold``, False if at least one does, or None
+            if no frequency data is available.
+        """
+        frequencies = self.get_frequencies()
+
+        if frequencies is None:
+            return None
+
+        return all(freq >= -threshold for freq in frequencies.values())
```
timmyte: Using the `threshold` parameter with `get_frequencies()`, this could be turned into:
```
return bool(frequencies)
```

src/opi/output/core.py
```diff
@@ -2198,6 +2198,96 @@ def get_free_energy_delta(self, *, index: int = -1) -> StrictFiniteFloat | None:
...
+        is_transition_state : bool | None
+            True if exactly one frequency lies below ``-threshold``, False otherwise, or None
+            if no frequency data is available.
+        """
+        frequencies = self.get_frequencies()
+
+        if frequencies is None:
+            return None
+
+        return sum(1 for freq in frequencies.values() if freq < -threshold) == 1
```
timmyte: Can't this be simplified using `self.get_imaginary_frequencies()` as the length of the return dict should be 1!?

tests/examples/test_exmp056_optts_freq.py
```diff
@@ -0,0 +1,30 @@
...
+@pytest.mark.examples
+@pytest.mark.orca
+def test_exmp056_optts_freq(example_input_file, tmp_path) -> None:
+    """Ensure OPTTS + NUMFREQ example runs and identifies the structure as a transition state."""
+    input_file = example_input_file(run_exmp056)
+    structure = Structure.from_xyz(input_file)
+
+    output = run_exmp056(structure=structure, working_dir=tmp_path)
+
+    assert output.get_final_energy() < 0
```
timmyte, line 17: Why would the energy ever be positive? Wouldn't this be a bug in ORCA?
Please add a comment.

timmyte (approved):
Thanks for adding these getters.
Please consider adding the `threshold` parameter consistently to all the new methods.

src/opi/output/core.py
haneug: To be more verbose and make clear what is checked I use `get_imaginary_frequencies` and do
```
return not bool(imaginary_frequencies)
```

src/opi/output/core.py
haneug: I agree, now I just check that the length is 1.

src/opi/output/core.py
haneug: get_imaginary_frequencies now becomes almost redundant but I kept it since I think it is still useful.

PR #248, opened 2026-06-02
feat: allow invalid output fields to be dropped instead of raising

haneug:
## Closes Issues
Closes None

## Description
- Adds an optional keyword `strict=True` to `Output.__init__()` and to `Output.parse()` which allows to drop the fields where Pydantic validation fail when it is set to `False`. This way we do not switch the current behavior but have the option for a less strict output generation. 

---

### Added 

- `Output.parse()`, `Output.parse_property()`, `Output.parse_gbw()`, and `Output.__init__()` now accept a `strict` parameter (default True). When set to False, output fields that fail Pydantic validation are silently set to None and a UserWarning is emitted instead of raising a ValidationError (#248).

### DIFF
--- a/src/opi/output/core.py
+++ b/src/opi/output/core.py
@@ -104,6 +104,7 @@ class Output:
         working_dir: Path | None = None,
         version_check: bool = True,
         parse: bool = False,
+        strict: bool = True,
     ):
         """
         ORCA output parser that is mainly based on the JSON-property and JSON-GBW file.
@@ -124,6 +125,10 @@ class Output:
             True: Create (if turned on by `create_gbw_json/create_property_json`) and parse JSONs files at the end of the initialization.
             False: Only return an Output object. In order to use the object to access the JSON data,
                    `Output.parse()` has to be called first.
+        strict: bool, default: True
+            If True, a ValidationError is raised when any field in the parsed output cannot be validated.
+            If False, fields that fail validation are silently set to None and a UserWarning is emitted.
+            Only takes effect when `parse=True`; otherwise pass `strict` directly to `Output.parse()`.
 
         Raises
         ----------
@@ -157,7 +162,7 @@ class Output:
 
         # // CREATE AND PARSE JSONS FILES
         if parse:
-            self.parse()
+            self.parse(strict=strict)
 
     def parse(
         self,
@@ -165,6 +170,7 @@ class Output:
         do_create_gbw_json: bool | None = None,
         read_prop_json: bool = True,
         read_gbw_json: bool = True,
+        strict: bool = True,
     ) -> None:
         """
         Create and read property- and gbw-JSON file(s) (according to `do_create_property_json` and `do_create_gbw_json`).
@@ -185,6 +191,9 @@ class Output:
             Whether to read the gbw JSON files. If True, the base gbw JSON file and any gbw JSON files from multi gbw
             runs (e.g., scan or neb) will be read. If False, none of these files will be read.
             If any of the JSON files that should be read are not present, a FileNotFoundError is raised.
+        strict: bool, default: True
+            If True, a ValidationError is raised when any field in the parsed output cannot be validated.
+            If False, fields that fail validation are silently set to None and a UserWarning is emitted.
 
         Raises
         ----------
@@ -192,20 +201,22 @@ class Output:
             If any JSON file should be read that is not present.
         """
         if read_prop_json:
-            self.parse_property(do_create_property_json=do_create_property_json)
+            self.parse_property(do_create_property_json=do_create_property_json, strict=strict)
         else:
             # // version check is performed with property JSON
             if self.do_version_check:
                 warn("No version check possible.")
 
         if read_gbw_json:
-            self.parse_gbw(do_create_gbw_json=do_create_gbw_json)
+            self.parse_gbw(do_create_gbw_json=do_create_gbw_json, strict=strict)
 
         # > Redump JSON files
         if self.do_redump_jsons:
             self._redump_jsons()
 
-    def parse_property(self, do_create_property_json: bool | None = None) -> None:
+    def parse_property(
+        self, do_create_property_json: bool | None = None, strict: bool = True
+    ) -> None:
         """
         Create and read property-JSON file.
 
@@ -214,6 +225,9 @@ class Output:
         do_create_property_json: bool | None, default: None
             Whether to create the property JSON file. If None, the file is only created if it is missing. If True,
             the existing file will be overwritten. If False, the file will not be created. Default is None.
+        strict: bool, default: True
+            If True, a ValidationError is raised when any field cannot be validated.
+            If False, fields that fail validation are set to None and a UserWarning is emitted.
         """
         # // Use default name if None was supplied
         if not self.property_json_file:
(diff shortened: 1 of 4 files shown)


src/opi/output/models/base/get_item.py
```diff
@@ -32,6 +34,34 @@ def get_clean_type_name(t: Any) -> str:
...
+        data: Any,
+        handler: ModelWrapValidatorHandler["GetItem"],
+        info: ValidationInfo,
+    ) -> "GetItem":
+        # > First try to run core Pydantic validation
+        try:
+            return handler(data)
+        except ValidationError as e:
+            # > Default is strict is set to True and the error is just raised
+            if (info.context or {}).get("strict", True):
```
timmyte: 1) Please add comments, why are you even looking for `strict` here.
Please explain that you are looking for the method parameter.
2) I would also add `strict` as class variable, instead of hard-coding the default here. 
So the expression turns into:
```
(info.context or {}).get("strict", cls.strict)
```

src/opi/output/models/base/get_item.py
```diff
@@ -32,6 +34,34 @@ def get_clean_type_name(t: Any) -> str:
...
+                raise
+            if not isinstance(data, dict):
+                raise
```
timmyte, line 67: Why is the exception also raise under this condition? Please extend the code documentation here.

src/opi/output/models/base/get_item.py
```diff
@@ -32,6 +34,34 @@ def get_clean_type_name(t: Any) -> str:
...
+            bad_keys = {str(error["loc"][0]) for error in e.errors() if error["loc"]}
```
timmyte, line 69: What are `bad_keys`? Please add comment describing the need to check this condition here as well.

src/opi/output/models/base/get_item.py
```diff
@@ -32,6 +34,34 @@ def get_clean_type_name(t: Any) -> str:
...
+            if not bad_keys:
+                raise
+            cleaned = {k: v for k, v in data.items() if str(k) not in bad_keys}
```
timmyte, line 75: What is cleaned?
Please add comments.

src/opi/output/models/base/get_item.py
```diff
@@ -32,6 +34,34 @@ def get_clean_type_name(t: Any) -> str:
...
+            result = handler(cleaned)  # re-raises if required fields are missing
+            warnings.warn(
+                f"{cls.__name__}: dropped fields due to validation errors: {bad_keys}",
+                stacklevel=2,
```
timmyte: Why explicitly `stacklevel=2`?
Please document this in the code.

src/opi/output/core.py
```diff
@@ -124,6 +125,10 @@ def __init__(
             True: Create (if turned on by `create_gbw_json/create_property_json`) and parse JSONs files at the end of the initialization.
             False: Only return an Output object. In order to use the object to access the JSON data,
                    `Output.parse()` has to be called first.
+        strict: bool, default: True
+            If True, a ValidationError is raised when any field in the parsed output cannot be validated.
+            If False, fields that fail validation are silently set to None and a UserWarning is emitted.
+            Only takes effect when `parse=True`; otherwise pass `strict` directly to `Output.parse()`.
```
timmyte, line 131:
```suggestion
        strict: bool, default: True
            If `True`, a `ValidationError` is raised when any field in the parsed output cannot be validated.
            If `False`, fields that fail validation are silently set to `None` and a `UserWarning` is emitted.
            Only takes effect when `parse=True`; otherwise pass `strict` directly to `Output.parse()`.
```

src/opi/output/core.py
```diff
@@ -185,27 +191,32 @@ def parse(
             Whether to read the gbw JSON files. If True, the base gbw JSON file and any gbw JSON files from multi gbw
             runs (e.g., scan or neb) will be read. If False, none of these files will be read.
             If any of the JSON files that should be read are not present, a FileNotFoundError is raised.
+        strict: bool, default: True
+            If True, a ValidationError is raised when any field in the parsed output cannot be validated.
+            If False, fields that fail validation are silently set to None and a UserWarning is emitted.
```
```suggestion
        strict: bool, default: True
            If `True`, a `ValidationError` is raised when any field in the parsed output cannot be validated.
            If `False`, fields that fail validation are silently set to `None` and a `UserWarning` is emitted.
```

src/opi/output/core.py
```diff
@@ -214,6 +225,9 @@ def parse_property(self, do_create_property_json: bool | None = None) -> None:
         do_create_property_json: bool | None, default: None
             Whether to create the property JSON file. If None, the file is only created if it is missing. If True,
             the existing file will be overwritten. If False, the file will not be created. Default is None.
+        strict: bool, default: True
+            If True, a ValidationError is raised when any field cannot be validated.
+            If False, fields that fail validation are set to None and a UserWarning is emitted.
```
```suggestion
            If `True`, a `ValidationError` is raised when any field cannot be validated.
            If `False`, fields that fail validation are set to `None` and a `UserWarning` is emitted.
```

src/opi/output/core.py
```diff
@@ -240,6 +256,9 @@ def parse_gbw(self, do_create_gbw_json: bool | None = None) -> None:
         do_create_gbw_json: bool | None, default: None
             Whether to create the gbw JSON files. If None, the files are only created if they are missing. If True,
             the existing files will be overwritten. If False, the files will not be created. Default is None.
+        strict: bool, default: True
+            If True, a ValidationError is raised when any field cannot be validated.
+            If False, fields that fail validation are set to None and a UserWarning is emitted.
```
```suggestion

            If `True`, a `ValidationError` is raised when any field cannot be validated.
            If `False`, fields that fail validation are set to `None` and a `UserWarning` is emitted.
```

tests/unit/test_output_partial_validation.py
```diff
@@ -0,0 +1,57 @@
...
+            warnings.simplefilter("always")
+            result = PropertyResults.model_validate(
+                {"calculation_info": {"charge": "bad", "mult": 1}},
+                context={"strict": False},
+            )
+
+        assert result.calculation_info is not None
+        assert result.calculation_info.charge is None
+        assert result.calculation_info.mult == 1
+        assert any("charge" in str(w.message) for w in caught)
```
timmyte, line 61: Please document the test mechanism with `charge` more extensively. It's not trivial. Thanks!

timmyte (changes requested):
Thanks for adding this feature. That's a very good idea.

It would be great if you could generally document your code more extensively, as it does introduce new concepts, which are not trivial.

What is not clear to me: You are saying, that in non-strict mode invalid keys are set to `None`? But what if their default value is not `None`?

src/opi/output/models/base/get_item.py
haneug: `stacklevel=2` is a bit arbitrary here so I removed it. Finding the caller outside of pydantic is a bit more involved and not worth it since the message is already descriptive enough.

src/opi/output/models/base/get_item.py
haneug: I added it as a class variable. To not make it a pydantic field and have typing I needed to use `ClassVar`:
```
    strict: ClassVar[bool] = (
        True  # > Annotated but with ClassVar so that it is not treated as a Pydantic field
    )
   ```

timmyte (approved):

PR #253, opened 2026-06-19
feat: add get_fcidump to output

haneug:
## Closes Issues
Closes None

## Description
- Added a small helper class that parses the FCIDUMP file and is able to return the important quantities (one and two-electron integrals) as numpy arrays. This object is populated and returned by `output.get_fcidump`.

---

### Added 

- Added `Output.get_fcidump()` and the `Fcidump` class for parsing FCIDUMP files generated by ORCA (#253)


src/opi/output/core.py
```diff
@@ -2742,3 +2743,25 @@ def get_free_solvation_energy_opencosmors(self) -> float | None:
         except (KeyError, IndexError):
             return None
         return float(free_solv_energy)
+
+    def get_fcidump(self) -> Fcidump | None:
+        """
+        Parse the fcidump file generated by ORCA and return its data in the `Fcidump` data class.
+        The fcidump file has to be generated by the ORCA job and cannot be generated on-the-fly after the calculation.
+        ORCA can generate fcidump files via the `dumpactints true` flag in the `%output` block.
```
```suggestion
        To generate FCIDUMP files, set the following options:
        ```
        %output
           dumpactints true
        end 
        ```
```
IMO this reads more concise.

src/opi/output/core.py
```diff
@@ -2742,3 +2743,25 @@ def get_free_solvation_energy_opencosmors(self) -> float | None:
         except (KeyError, IndexError):
             return None
         return float(free_solv_energy)
+
+    def get_fcidump(self) -> Fcidump | None:
+        """
+        Parse the fcidump file generated by ORCA and return its data in the `Fcidump` data class.
```
```suggestion
        Parse the FCIDUMP file generated by ORCA and return its data in the `Fcidump` data class.
```

src/opi/output/core.py
```diff
@@ -2742,3 +2743,25 @@ def get_free_solvation_energy_opencosmors(self) -> float | None:
...
+        The fcidump file has to be generated by the ORCA job and cannot be generated on-the-fly after the calculation.
```
```suggestion
        The FCIDUMP file has to be generated by the ORCA job and cannot be generated on-the-fly after the calculation.
```

src/opi/output/core.py
```diff
@@ -2742,3 +2743,25 @@ def get_free_solvation_energy_opencosmors(self) -> float | None:
...
+        ORCA can generate fcidump files via the `dumpactints true` flag in the `%output` block.
+        See `BlockOutput` for the related block option.
+
+        Returns
+        -------
+        fcidump_data: Fcidump | None
+            The parsed fcidump data or None if the file is not present or could not be parsed.
```
```suggestion
            The parsed FCIDUMP data or None if the file is not present or could not be parsed.
```

src/opi/output/fcidump.py
```diff
@@ -0,0 +1,161 @@
+"""Parse a potential FCIDUMP file"""
```
timmyte, line 1: A module that contains one primary class, should also be named after that class -> `fcidump.py`

src/opi/output/fcidump.py
```diff
@@ -0,0 +1,161 @@
...
+import re
+from dataclasses import dataclass, field
+from functools import cached_property
+from pathlib import Path
+
+import numpy as np
+
+
+@dataclass
+class Fcidump:
```
timmyte, line 13: According to Python's style guide:
In words that are fully or partially acronyms, the letters comprising the acronym are typically written in uppercase: `FCIDump`

Most of the OPI code treats an acronym as one single *syllable* and we only capitalize the first letter of each *syllable*. Which is fine.
But this is not done in all case. We should pay attention to this.

src/opi/output/fcidump.py
```diff
@@ -0,0 +1,161 @@
...
+
+        Raises
+        -------
+        ValueError
+            If the FCIDUMP file cannot be parsed.
+        FileNotFoundError
+            If the FCIDUMP file cannot be found at the given path.
+        """
+
+        if isinstance(path, str):
```
timmyte: Just do `path = Path(path)`. 
This also takes care if `path` is of type: `os.PathLike` (which is unlikely).

Doing `Path(Path())` just returns the inner path. No harm done.

src/opi/output/fcidump.py
```diff
@@ -0,0 +1,161 @@
...
+            path = Path(path)
+
+        if not path.is_file():
+            raise FileNotFoundError(f"{cls.__name__}: FCIDUMP file not found at {path}")
+
+        with open(path) as f:
+            text = f.read()
```
timmyte: 1) `text = path.read_text()` is even shorter ;)
2) I understands it's tempting to read file in one go. Makes handling easier. But this can be become a bottleneck if the files grow very large.
Did you consider this?

src/opi/output/fcidump_parser.py
```diff
@@ -0,0 +1,161 @@
...
+                (c, d, a, b),
+                (d, c, a, b),
+                (c, d, b, a),
+                (d, c, b, a),
+            ]:
+                tensor[p, q, r, s] = val
+        return tensor
+
+    @classmethod
+    def parse_fcidump(cls, path: Path | str) -> "Fcidump":
```
timmyte: 1) IMO `from_file()` is a more accessible name than `parse_fcidump()`.
2) Are the FCIDUMP files documented anywhere? Could add the link to the docstring, if so.

src/opi/output/fcidump.py
```diff
@@ -0,0 +1,161 @@
...
+            # > Two-electron tensor
+            else:
+                dump.two_electron[(i, j, k, ll)] = val
+
+        return dump
+
+    @classmethod
+    def _get_int(cls, key: str, header: str) -> int:
+        """Return the integer value of the given key."""
+        m = re.search(rf"{key}\s*=\s*(\d+)", header, re.IGNORECASE)
```
timmyte, line 167: What about negative numbers?
If these lines always look as follows:
```
key    =     1234   [END OF LINE]
```
You could use `.partition("=")`

src/opi/output/fcidump_parser.py
```diff
@@ -0,0 +1,161 @@
...
+        """Return the integer value of the given key."""
+        m = re.search(rf"{key}\s*=\s*(\d+)", header, re.IGNORECASE)
+        if m is None:
+            raise ValueError(f"{cls.__name__}: Could not parse {key}")
+        return int(m.group(1))
+
+    @classmethod
+    def _get_int_list(cls, key: str, header: str) -> list[int]:
+        """Return a list of integers corresponding to the given key."""
+        m = re.search(rf"{key}\s*=\s*([\d,\s]+)", header, re.IGNORECASE)
```
timmyte:
```suggestion
        m = re.search(rf"{key}\s*=\s*(\d+(\s*,\s*\d+)+)", header, re.IGNORECASE)
```
What about a more precise regex!?
This still does not account for leading plus/minus symbol.

tests/unit/test_output_fcidump_parser.py
```diff
@@ -0,0 +1,79 @@
+import textwrap
```
timmyte: Please add units test for `_get_int_list()` and `_get_int()` specifically.

tests/unit/test_output_fcidump_parser.py
```diff
@@ -0,0 +1,79 @@
...
+from pathlib import Path
+
+import pytest
+
+from opi.output.fcidump_parser import Fcidump
+
+
+@pytest.mark.unit
+def test_parse_fcidump_header(tmp_path: Path) -> None:
+    fcidump_text = textwrap.dedent("""\
```
timmyte: I would actually not remove the indentation.
If a file format does not rely on indentation, then our file parser shouldn't as well. And the current also implementation does (good implementation 👍)
So I would actively test that.

timmyte (changes requested):

src/opi/output/fcidump.py
haneug: Negative numbers would not be valid in the cases where we use `_get_int`. I should probably add that to the docstring

src/opi/output/fcidump_parser.py
haneug: For now I do not see a problem with plus/minus symbols as there should be none in a valid FCIDUMP file

src/opi/output/fcidump.py
haneug: 1. Thanks
2. I did think about this. FCIDUMP is a flat text format with no index, so there is no way to read part of it without scanning the whole file anyway, so there is not much to gain by lazy reading it. Also if you use the data for RHF/UHF/FCI you typically need all of it anyway.
The actual bottleneck is the generation of the numpy arrays for the two-electron integrals. The most expensive part is unsymmetrizing them, since FCIDUMP only stores symmetry-unique entries and building the dense tensor means writing each value to 8 positions. Instead of the for loop I now use a vectorized approach which should be faster.

src/opi/output/core.py
```diff
@@ -2742,3 +2743,33 @@ def get_free_solvation_energy_opencosmors(self) -> float | None:
         except (KeyError, IndexError):
             return None
         return float(free_solv_energy)
+
+    def get_fcidump(self) -> Fcidump | None:
+        """
+        Parse the FCIDUMP file generated by ORCA and return its data in the `Fcidump` data class.
```
timmyte, line 2749: Nicely documented!!

src/opi/output/fcidump.py
```diff
@@ -0,0 +1,183 @@
...
+            (c, d, a, b),
+            (d, c, a, b),
+            (c, d, b, a),
+            (d, c, b, a),
+        ):
+            tensor[p, q, r, s] = vals
+        return tensor
+
+    @classmethod
+    def from_file(cls, path: Path | str) -> "Fcidump":
```
timmyte:
```suggestion
    def from_file(cls, path: Path | str) -> Self:
```

src/opi/output/fcidump.py
```diff
@@ -0,0 +1,183 @@
...
+    orbsym: list[int]
+    isym: int
+    one_electron: dict[tuple[int, int], float] = field(default_factory=dict)
+    two_electron: dict[tuple[int, int, int, int], float] = field(default_factory=dict)
+    orbital_energies: dict[int, float] = field(default_factory=dict)
+    e_nuc: float = 0.0
+    path: Path = field(default_factory=Path)
+
+    @cached_property
+    def hcore_matrix(self) -> np.ndarray:
```
timmyte: Please don't use generic return type annotation here. You can fill in the unknowns: `np.ndarray[tuple[int, int], np.dtype[np.float]]`

Also functions that return Numpy arrays.

tests/unit/test_output_fcidump_parser.py
```diff
@@ -0,0 +1,146 @@
...
+    assert dump.ms2 == 0
+    assert dump.orbsym == [1, 1]
+    assert dump.isym == 0
+    assert pytest.approx(dump.e_nuc) == -1.2345678901
+
+
+@pytest.mark.unit
+@pytest.mark.output
+def test_get_int() -> None:
+    """`_get_int` extracts integer header values, tolerating whitespace and mixed-case keys."""
```
timmyte: I would always end function references with empty parenthesis: `_get_int()` 
Makes difference between functions and other objects clearer ;)

Also check the docstrings below.

timmyte (approved):
Looks good.
Just some minor annotations to fix ;)

PR #255, opened 2026-07-01
WIP:feat/notebook tests

### LINKED ISSUE #202: Run tutorial notebooks as part of the tests
In order to verify that the notebooks work with the current OPI version we should add their execution to the testsuite.

Separate notebook tests that verify that they run without any errors.

nakul680:
## Closes Issues
Closes #202

## Description
- Adds a test script that runs all tutorial notebooks( with a few exceptions of notebooks that rely on external packages) and checks whether these notebooks are executed without issues.

---

## Release Notes

### Added 

- Add `test_notebooks()`, which will execute all tutorial notebooks.(#255)
- Add `notebooks` pytest tag for `test_notebooks()`(#255)

### DIFF
--- a/tests/test_notebook.py
+++ b/tests/test_notebook.py
@@ -1,0 +1,64 @@
+#!/usr/bin/env python
+"""Run all notebooks in tests/notebooks/ and report pass/fail."""
+
+import subprocess
+import sys
+import tempfile
+import time
+from pathlib import Path
+
+import pytest
+
+# NOTEBOOKS_DIR = Path(__file__).parent.parent / "docs/contents/notebooks"
+NOTEBOOKS_DIR = Path(__file__).parent / "notebooks"
+
+
+def run_notebook(nb: Path) -> tuple[bool, float, str]:
+    start = time.monotonic()
+    with tempfile.TemporaryDirectory() as tmp:
+        result = subprocess.run(
+            [
+                "jupyter",
+                "nbconvert",
+                "--execute",
+                "--to",
+                "notebook",
+                "--ExecutePreprocessor.timeout=1800",
+                f"--ExecutePreprocessor.kernel_cwd={tmp}",
+                "--output",
+                nb.name,
+                str(nb),
+            ],
+            capture_output=True,
+            text=True,
+            cwd=nb.parent,
+        )
+    elapsed = time.monotonic() - start
+    return result.returncode == 0, elapsed, result.stdout + result.stderr
+
+
+@pytest.mark.notebooks
+def test_notebooks() -> None:
+    notebooks = sorted(NOTEBOOKS_DIR.glob("*.ipynb"))
+    if not notebooks:
+        print(f"No notebooks found in {NOTEBOOKS_DIR}")
+        sys.exit(1)
+
+    results: list[tuple[str, bool, float]] = []
+    for nb in notebooks:
+        print(f"  running {nb.name} ...", end="", flush=True)
+        ok, elapsed, output = run_notebook(nb)
+        status = "PASS" if ok else "FAIL"
+        print(f" {status} ({elapsed:.1f}s)")
+        if not ok:
+            print(output)
+        results.append((nb.name, ok, elapsed))
+
+    print()
+    failures = [name for name, ok, _ in results if not ok]
+    print(f"{len(notebooks) - len(failures)}/{len(notebooks)} notebooks passed")
+    if failures:
+        print("Failed:")
+        for name in failures:
+            print(f"  - {name}")
+        sys.exit(1)
--- a/tests/test_notebook.py
+++ b/tests/test_notebook.py
@@ -9,8 +9,7 @@ from pathlib import Path
 
 import pytest
 
-# NOTEBOOKS_DIR = Path(__file__).parent.parent / "docs/contents/notebooks"
-NOTEBOOKS_DIR = Path(__file__).parent / "notebooks"
+NOTEBOOKS_DIR = Path(__file__).parent.parent / "docs/contents/notebooks"
 
 
 def run_notebook(nb: Path) -> tuple[bool, float, str]:
@@ -27,6 +26,8 @@ def run_notebook(nb: Path) -> tuple[bool, float, str]:
                 f"--ExecutePreprocessor.kernel_cwd={tmp}",
                 "--output",
                 nb.name,
+                "--output-dir",
+                tmp,
                 str(nb),
             ],
             capture_output=True,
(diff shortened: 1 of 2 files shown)


tests/test_notebook.py
```diff
@@ -0,0 +1,81 @@
...
+from pathlib import Path
+
+import pytest
+
+NOTEBOOKS_DIR = Path(__file__).parent.parent / "docs/contents/notebooks"
+BLACKLIST = ["extopt", "ir_spectrum", "ml_properties"]
+
+
+def run_notebook(nb: Path) -> tuple[bool, float, str]:
+    """
```
timmyte, line 18: Please complete the docstring.

tests/test_notebook.py
```diff
@@ -0,0 +1,81 @@
...
+    Parameters
+    ----------
+    nb
+
+    Returns
+    -------
+
+    """
+    start = time.monotonic()
+    with tempfile.TemporaryDirectory() as tmp:
```
timmyte, line 29: Describe in detail what you are doing here:
* Where are notebooks executed?
* How are they executed?
* Why is there a timeout?
* Why does it need to copied over?

Where do make sure `jupyter` is installed?

tests/test_notebook.py
```diff
@@ -0,0 +1,81 @@
...
+            [
+                "jupyter",
+                "nbconvert",
+                "--execute",
+                "--to",
+                "notebook",
+                "--ExecutePreprocessor.timeout=1800",
+                "--output",
+                nb.name,
+                "--output-dir",
```
timmyte, line 42: I guess `--output-dir` is redundant, as `cwd=tmp`.

tests/test_notebook.py
```diff
@@ -0,0 +1,81 @@
...
+    return result.returncode == 0, elapsed, result.stdout + result.stderr
+
+
+@pytest.mark.notebooks
+def test_notebooks() -> None:
+    """
+    Function to iteratively test whether all notebooks (with exception to the ones to be excluded, which are defined in `BlACKLIST`) execute
+    without issues. In the case that a notebook execution fails,the function will print out which notebooks fail and the error message.
+    """
+    notebooks = [nb for nb in NOTEBOOKS_DIR.glob("*.ipynb") if nb.stem not in BLACKLIST]
```
timmyte, line 60: I think this should actually be a fixture.

Instead of one *master* function like here that acts as whole notebooks testsuite, I would boil this down to a single function, that takes a notebook, executes it and returns a flag if the notebook failed or not.

I would also keep the reporting to a minimum. Pytests captures the output anyway.
Though we should keep the printing of the timings. This is good for debugging and make sure out testsuite does not take too long ;)

I think @haneug already wrote a custom summary report for one type of tests that we have. Maybe he can help.

tests/test_notebook.py
```diff
@@ -0,0 +1,81 @@
...
+                "--output-dir",
+                tmp,
+                str(nb_copy),
+            ],
+            capture_output=True,
+            text=True,
+            cwd=tmp,
+        )
+    elapsed = time.monotonic() - start
+    return result.returncode == 0, elapsed, result.stdout + result.stderr
```
timmyte, line 51: Good idea with the boolean!

tests/test_notebook.py
```diff
@@ -0,0 +1,81 @@
...
+import subprocess
+import sys
+import tempfile
+import time
+from pathlib import Path
+
+import pytest
+
+NOTEBOOKS_DIR = Path(__file__).parent.parent / "docs/contents/notebooks"
+BLACKLIST = ["extopt", "ir_spectrum", "ml_properties"]
```
timmyte, line 14: Why are they blacklisted?
Please add a comment.

tests/test_notebook.py
```diff
@@ -0,0 +1,81 @@
...
+import shutil
+import subprocess
+import sys
+import tempfile
+import time
+from pathlib import Path
+
+import pytest
+
+NOTEBOOKS_DIR = Path(__file__).parent.parent / "docs/contents/notebooks"
```
timmyte, line 13: Instead of hardcoding these complex paths, I think we should start adding symlinks to the repo. If we refactor the layout at some point, finding broken symlinks, is a peace of cake, but correcting the hardcoded paths is painful.

@haneug 
Objections?

timmyte (changes requested):
Looks good. It's definitely about time we also test the notebooks.
We should just polish the architecture a little.

tests/test_notebook.py
haneug: No, sounds good!

tests/test_notebook.py
haneug: Because they require additional external dependencies that cannot easily be set up by hand (extopt and ml_properties at least). This should be documented here. About ir_spectrum I am not sure?

tests/test_notebook.py
```diff
@@ -0,0 +1,81 @@
...
+            ],
+            capture_output=True,
+            text=True,
+            cwd=tmp,
+        )
+    elapsed = time.monotonic() - start
+    return result.returncode == 0, elapsed, result.stdout + result.stderr
+
+
+@pytest.mark.notebooks
```
haneug, line 54: notebooks also always run orca so I guess `@pytest.mark.orca` is also appropriate here

tests/test_notebook.py
haneug: I agree with the fixture / parametrize. I did add a custom error message when example tests fail although I do not remember much of this I am happy to discuss this.

tests/test_notebook.py
```diff
@@ -0,0 +1,81 @@
...
+def test_notebooks() -> None:
+    """
+    Function to iteratively test whether all notebooks (with exception to the ones to be excluded, which are defined in `BlACKLIST`) execute
```
haneug, line 57: Typo in `BlACKLIST`

tests/test_notebook.py
```diff
@@ -0,0 +1,81 @@
...
+    without issues. In the case that a notebook execution fails,the function will print out which notebooks fail and the error message.
+    """
+    notebooks = [nb for nb in NOTEBOOKS_DIR.glob("*.ipynb") if nb.stem not in BLACKLIST]
+    if not notebooks:
+        print(f"No valid notebooks found in {NOTEBOOKS_DIR}")
+        sys.exit(1)
```
haneug, line 63: use `pytest.fail()` instead

haneug (changes requested):

PR #261, opened 2026-07-10
feat: add type checking for `ncores` and `memory` setters

### LINKED ISSUE #259: Add type check for setters for `ncores` and `memory`
Add type check for the setters for `ncores` and `memory` properties

nakul680:
## Closes Issues
Closes #259 

## Description
- Add type checks to the setters of `ncores` and `memory` properties in `Input`. Additionally it will be checked if the value given is positive.
- Add unit tests to ensure no invalid values are accepted.

### Added 

- Add type checks for `ncores` and `memory` setters in and checks for invalid negative values.(#261)


tests/unit/test_input_core.py
```diff
@@ -0,0 +1,66 @@
...
+@pytest.fixture
+def empty_calc():
+    """An empty instance of `Calculator`."""
+    empty_calc = Calculator("test", version_check=False)
+    return empty_calc
+
+
+@pytest.mark.unit
+@pytest.mark.input
+@pytest.mark.parametrize("ncores", [0, 4, None])
```
haneug, line 19: I was quite surprised that 0 is actually a valid number but ORCA just runs with it

src/opi/input/core.py
```diff
@@ -118,8 +118,11 @@ def ncores(self, value: int | None) -> None:
         ----------
         value : int | None
         """
-        if value is not None and value < 0:
-            raise ValueError(f"{self.__class__.__name__}.ncores must be a positive integer.")
+        if value is not None:
+            if not isinstance(value, int):
+                raise TypeError(f"{self.__class__.__name__}.ncores must be an integer.")
+            elif value < 0:
```
haneug, line 134: Please document in the doc string that the integer should be Nonnegative

src/opi/input/core.py
```diff
@@ -134,8 +137,11 @@ def memory(self, value: int | None) -> None:
         ----------
         value : int | None
         """
-        if value is not None and value < 0:
-            raise ValueError(f"{self.__class__.__name__}.memory must be a positive integer.")
+        if value is not None:
+            if not isinstance(value, int):
+                raise TypeError(f"{self.__class__.__name__}.memory must be an integer.")
+            elif value < 0:
```
haneug, line 163: Please document in the doc string that the integer should be Nonnegative

haneug (changes requested):
Thanks for the PR. Please correct the issue number in the Change log section and add it to the actual `CHANGELOG.md` file 
Edit: Also the Change log section should only be one thing in Added. Something like Added: type checks for ncore and memory. That the feature comes with tests does not need to be mentioned in the Change-log.

haneug (changes requested):
Please add the release notes to the `CHANGELOG.md` file. It seems that your release note is incomplete. Also remove the dots `...` below Closes, the `(Project adheres to Keep a Changelog; Every entry should start in upper-case and end with (#<pr-id>); Remove sections that don't apply)`, and the empty point in the description of your PR.

haneug (approved):

PR #262, opened 2026-07-10
feat: add cleanup routines for files in working directory

### LINKED ISSUE #258: Add cleanup functions for files created by ORCA
Create cleanup functions such as : 
- `Output.cleanup(basename)` : To cleanup all files created by the ORCA job within the working directory.
- `Output.remove_temp_files()` : To clean up all temporary files created by ORCA within the working directory.
- `Output.remove_working_dir()`: To remove the entire working directory

nakul680:
## Closes Issues
Closes #258 

## Description
- Add functions to cleanup files created by ORCA jobs.

---

### Added 

- Add `functions to clean up files created by ORCA jobs.(#262)


src/opi/output/core.py
```diff
@@ -536,6 +513,53 @@ def get_file(self, suffix: str, /, *, basename: str = "") -> Path:
...
+
+        Parameters
+        ----------
+        basename : str | None, default: None
+            Basename of the job whose files should be deleted.
+            If not given, `self.basename` is used.
+        """
+        self._delete_files(basename)
+
+    def cleanup_temp_files(self, basename: str | None = None) -> None:
```
haneug, line 599: This will delete `basename.tmp` and `basename.proc` but these are not actual ORCA files. Actual ORCA files look like `basename.something.tmp` or `basename.something.tmp.0`. The suffixes need to be treated as glob patterns. Something like `working_dir.glob(f"{basename}*.tmp*"`.

src/opi/output/core.py
```diff
@@ -536,6 +513,53 @@ def get_file(self, suffix: str, /, *, basename: str = "") -> Path:
...
+        basename = basename if basename is not None else self.basename
+        files: Iterable[Path]
+        if suffixes is None:
+            files = self.working_dir.glob(f"{basename}*")
+        else:
+            files = (self.get_file(suffix, basename=basename) for suffix in suffixes)
+        for file in files:
+            file.unlink(missing_ok=True)
+
+    def cleanup_files(self, basename: str | None = None) -> None:
```
haneug, line 585: `cleanup_files("")` will delete the whole working dir since the glob will become `*` You should use a truthiness check for the basename in `_delete_files` to be consistent.

src/opi/output/core.py
```diff
@@ -536,6 +513,53 @@ def get_file(self, suffix: str, /, *, basename: str = "") -> Path:
...
+            Basename of the job whose files should be deleted.
+            If not given, `self.basename` is used.
+        suffixes : Sequence[str] | None, default: None
+            If given, only files with these suffixes are deleted.
+            If None, all files matching the basename are deleted.
+        """
+        basename = basename if basename is not None else self.basename
+        files: Iterable[Path]
+        if suffixes is None:
+            files = self.working_dir.glob(f"{basename}*")
```
haneug, line 571: This is very general as it will also delete for example `f"{basename}_*". This is generality is required since there are many different ORCA files we want to delete, but you should warn about this in the docstring.

src/opi/output/core.py
```diff
@@ -536,6 +513,55 @@ def get_file(self, suffix: str, /, *, basename: str = "") -> Path:
...
+        basename = basename if basename is not None else self.basename
+        if basename is None:
+            raise ValueError("No basename specified")
+        files: Iterable[Path]
+        if suffixes is None:
+            files = self.working_dir.glob(f"{basename}*")
+        else:
+            files = (self.get_file(suffix, basename=basename) for suffix in suffixes)
+        for file in files:
+            file.unlink(missing_ok=True)
```
haneug: will crash if file is actually a dir. Please check with is_file() before deleting.

haneug (changes requested):
Nice PR. Please see my comments. It would also be good to add some unit tests to see which files get deleted.

src/opi/output/core.py
```diff
@@ -536,6 +536,63 @@ def get_file(self, suffix: str, /, *, basename: str = "") -> Path:
             basename = self.basename
         return self.working_dir / (basename + suffix)
 
+    def _delete_files(self, basename: str | None, *, suffixes: Sequence[str] | None = None) -> None:
```
timmyte: 1) Any particular reason that `basename` has no default?
I would assume having one makes the API a bit smoother.
2) I'm note huge fan to allow string parameters to also be `None`, because an empty string typically acts as the `null` value. Any particular reason to have `None` here?
3) I would use an empty tuple as default for `Sequence`-like parameters.

src/opi/output/core.py
```diff
@@ -536,6 +536,63 @@ def get_file(self, suffix: str, /, *, basename: str = "") -> Path:
...
+        ----------
+        basename : str | None
+            Basename of the job whose files should be deleted.
+            If not given, `self.basename` is used.
+        suffixes : Sequence[str] | None, default: None
+            If given, treated as glob patterns matched against `{basename}*{suffix}`;
+            only matching files are deleted.
+            If None, all files matching the basename are deleted.
+        """
+        basename = basename if basename is not None else self.basename
```
timmyte: Building upon my empty string vs. `None` point above:
With string variables, I would always go for `if <string>` or `if not <string>`, this avoids common pitfalls.
```suggestion
        basename = basename if basename else self.basename
```

src/opi/output/core.py
```diff
@@ -536,6 +536,63 @@ def get_file(self, suffix: str, /, *, basename: str = "") -> Path:
...
+        if not basename:
+            raise ValueError("No basename specified")
+        files: Iterable[Path]
+        if suffixes is None:
```
timmyte: What if the sequence is empty? Which behavior is expected then?

src/opi/output/core.py
```diff
@@ -536,6 +536,63 @@ def get_file(self, suffix: str, /, *, basename: str = "") -> Path:
...
+            files = self.working_dir.glob(f"{basename}*")
+        else:
+            files = (
```
timmyte, line 574: Very good to use generators!!

src/opi/output/core.py
timmyte: @haneug 
Shouldn't we go for `BASENAME*.tmp.*` (which would require to also glob for another pattern: `BASENAME*.tmp`). I would assume `BASENAME*.tmp*` is a bit too broad!?

src/opi/output/core.py
```diff
@@ -536,6 +536,63 @@ def get_file(self, suffix: str, /, *, basename: str = "") -> Path:
...
+                file
+                for suffix in suffixes
+                for file in self.working_dir.glob(f"{basename}*{suffix}*")
```
timmyte: I would not automatically assume the second wildcard at the end. Thereby `suffix` is not a suffix anymore, but could also become an infix.

IMO it would be safer to glob for `"{basename}*{suffix}"`. And if one likes to have a wildcard at the end, it should come as part of the suffix.

src/opi/output/core.py
```diff
@@ -536,6 +536,63 @@ def get_file(self, suffix: str, /, *, basename: str = "") -> Path:
...
+        basename : str | None
+            Basename of the job whose files should be deleted.
+            If not given, `self.basename` is used.
+        suffixes : Sequence[str] | None, default: None
+            If given, treated as glob patterns matched against `{basename}*{suffix}`;
+            only matching files are deleted.
+            If None, all files matching the basename are deleted.
+        """
+        basename = basename if basename is not None else self.basename
+        if not basename:
```
timmyte, line 561: Good safeguard!!

timmyte (changes requested):
Thanks for implementing these helper functions.
There are some design decisions, which I would like to discuss.
@haneug 
Also happy for your input.

src/opi/output/core.py
haneug: I agree `BASENAME*.tmp.*` is the way to go

src/opi/output/core.py
nakul680: If the sequences is empty the function would essentially be a no-op

src/opi/output/core.py
haneug: and `BASENAME*.tmp`

src/opi/output/core.py
timmyte: @nakul680 
Could you please add a safeguard, to make sure `basename` does not contain path elements such as `../` or `./`.

src/opi/output/core.py
nakul680: I think having a None value makes it clear that the option of providing a basename other than self.basename is optional , and if its not given then OPI will default to self.basename. I also dont really think the API is changed much . I would like to maintain as much clarity as possible when it comes to deleting files which is why i added the None option.

I would however add a default of None

src/opi/output/core.py
nakul680: there is a safeguard that doesnt allow basename to have path separators

src/opi/output/core.py
```diff
@@ -536,6 +536,67 @@ def get_file(self, suffix: str, /, *, basename: str = "") -> Path:
...
+                for file in self.working_dir.glob(f"{basename}*{suffix}")
+            )
+        for file in files:
+            if file.is_file():
+                file.unlink(missing_ok=True)
+
+    def cleanup_files(self, basename: str | None = None) -> None:
+        """
+        Delete all files in `working_dir` belonging to the job with the given basename. This will also include files whose
+        names contain the basename, for example, if the basename is `job` and there exists files with basename 'job_1',
```
haneug: It will also delete files from jobs which have the same basename with additional suffixes such as `job_1` but not `example_job`. So I suggest: who have the same basename with additional labels such as job_1

src/opi/output/core.py
```diff
@@ -536,6 +536,67 @@ def get_file(self, suffix: str, /, *, basename: str = "") -> Path:
...
+        ----------
+        basename : str | None, default: None
+            Basename of the job whose files should be deleted.
+            If not given, `self.basename` is used.
+        """
+        self._delete_files(basename)
+
+    def cleanup_temp_files(self, basename: str | None = None) -> None:
+        """
+        Delete temporary files in `working_dir` belonging to the job with the given basename.
```
haneug: here you should add the same warning as in `clean_up_files()`

haneug (approved):
Looks good to me! Please update the docstrings of the `clean_up` functions and then we are good to go from my side.

src/opi/output/core.py
```diff
@@ -536,6 +536,67 @@ def get_file(self, suffix: str, /, *, basename: str = "") -> Path:
             basename = self.basename
         return self.working_dir / (basename + suffix)
 
+    def _delete_files(
+        self, basename: str | None = None, *, suffixes: Sequence[str] = tuple()
```
```suggestion
        self, basename: str | None = None, *, suffixes: Sequence[str] = ()
```

src/opi/output/core.py
```diff
@@ -536,6 +536,67 @@ def get_file(self, suffix: str, /, *, basename: str = "") -> Path:
...
+
+        Parameters
+        ----------
+        basename : str | None, default: None
+            Basename of the job whose files should be deleted.
+            If not given, `self.basename` is used.
+        suffixes : Sequence[str] | None, default: None
+            If given, treated as glob patterns matched against `{basename}*{suffix}`;
+            only matching files are deleted.
+            If None, all files matching the basename are deleted.
```
timmyte, line 552: Please document that the `ValueError`s in the `Raises` section.

src/opi/output/core.py
```diff
@@ -536,6 +536,67 @@ def get_file(self, suffix: str, /, *, basename: str = "") -> Path:
...
+        """
+        basename = basename if basename else self.basename
+        if not basename:
+            raise ValueError("No basename specified")
+        if "/" in basename or "\\" in basename:
```
timmyte: Please don't assume special characters in paths, use Python's intrinsic tools for such operation. These handle all the edge-case, that you would have to cover in a re-implementation yourself.

Tl;dr:
```
# > Making sure that `basename` is not a file path.
if len(Path(basename).parts) > 1: ...
```

src/opi/output/core.py
```diff
@@ -536,6 +536,67 @@ def get_file(self, suffix: str, /, *, basename: str = "") -> Path:
...
+            raise ValueError(f"Basename must not contain path separators: {basename!r}")
+        files: Iterable[Path]
```
timmyte, line 568: Please add a blank line above to separate input checks, from the actual cleanup and also some comments below to briefly communicate what the individual if, else and loops do.

tests/unit/test_output_cleanup.py
```diff
@@ -0,0 +1,108 @@
...
+- `Output._delete_files()`
+"""
+
+
+def _make_output(basename: str, working_dir: Path) -> Output:
+    output = Output(basename, working_dir=working_dir, version_check=False)
+    return output
+
+
+def _create_file(directory: Path, *names: str) -> None:
```
timmyte, line 22: Please add a docstring. Also include what's the purpose of the function.

tests/unit/test_output_cleanup.py
```diff
@@ -0,0 +1,108 @@
...
+Unit tests for Output cleanup functions.
+
+This module contains tests for the file-deletion helpers:
+- `Output.cleanup_files()`
+- `Output.cleanup_temp_files()`
+- `Output._delete_files()`
+"""
+
+
+def _make_output(basename: str, working_dir: Path) -> Output:
```
timmyte, line 17: Please add a docstring. Also include what's the purpose of the function.

timmyte (approved):
Thanks for implementing this.
This will be helpful. Please consider my final remarks below before merging.

PR #267, opened 2026-07-24
feat: fcidump writer and additional initializer from_arrays

haneug:
## Closes Issues
Closes None

## Description
- Add a function `to_file()` to the `Fcidump` class to write an FCIDUMP file. Also add new initializer `from_arrays()`.

---

### Added 

- Added `Fcidump.from_arrays()` for constructing a `Fcidump` from numpy arrays and `Fcidump.to_file()` for writing FCIDUMP files (#267).
- `Fcidump` now raises a `ValueError` if the one- or two-electron integrals contain symmetry-equivalent keys (#267).


src/opi/output/fcidump.py
```diff
@@ -1,6 +1,7 @@
-"""Parse a potential FCIDUMP file"""
+"""Read, create, and write FCIDUMP files"""
```
timmyte, line 1: Much better description ;)

src/opi/output/fcidump.py
```diff
@@ -15,6 +16,10 @@ class Fcidump:
     Reads and stores data from a FCIDUMP file. One and two-electrons integrals are stored as dicts and can be
     accessed as numpy arrays via `hcore_matrix` and `eri_tensor`.
 
+    Besides parsing a file with `from_file`, objects can be created directly from the integral
```
timmyte:
```suggestion
    Besides parsing a file with `from_file()`, objects can be created directly from the integral
```
It's good practice to always end with `()` when referring to functions or methods.
And I think we should also do this here.

Please also make sure to do this in the PR description, there you also refer to two functions.

src/opi/output/fcidump.py
```diff
@@ -88,6 +118,152 @@ def eri_tensor(self) -> np.ndarray[tuple[int, int, int, int], np.dtype[np.float6
...
+    @classmethod
+    def from_arrays(
+        cls,
+        hcore: np.ndarray[tuple[int, int], np.dtype[np.float64]],
+        eri: np.ndarray[tuple[int, int, int, int], np.dtype[np.float64]],
+        nelec: int,
+        ms2: int = 0,
+        *,
+        e_nuc: float = 0.0,
+        orbsym: Sequence[int] | None = None,
```
timmyte: I would use `()` as default instead of `None`. You can later still implicitly cover both cases with `if obrsym: ...`.

src/opi/output/fcidump.py
```diff
@@ -88,6 +118,152 @@ def eri_tensor(self) -> np.ndarray[tuple[int, int, int, int], np.dtype[np.float6
             tensor[p, q, r, s] = vals
         return tensor
 
+    @classmethod
+    def from_arrays(
+        cls,
+        hcore: np.ndarray[tuple[int, int], np.dtype[np.float64]],
```
timmyte: As you cast `hcore` and `eri` to arrays below any way, you could also allow `ArrayLike` here. You just would have to check the dimensions of the resulting `numpy` arrays then later.

tests/unit/test_output_fcidump_parser.py
```diff
@@ -144,3 +145,111 @@ def test_eri_tensor_shape_and_symmetry() -> None:
...
+@pytest.mark.unit
+@pytest.mark.output
+def test_hcore_matrix_conflicting_transposed_keys_stay_symmetric() -> None:
+    """With both (i,j) and (j,i) stored, the reconstructed matrix must still be symmetric."""
+    dump = Fcidump(
+        norb=2, nelec=2, ms2=0, orbsym=[1, 1], isym=0, one_electron={(1, 2): 0.1, (2, 1): 0.2}
+    )
+    mat = dump.hcore_matrix
+
+    assert pytest.approx(mat[0, 1]) == mat[1, 0]
```
timmyte: I wouldn't base the test on a single, fixed value. That this works for a 2x2-matrix could just be a coincidence.
I would vary the dimensions of the matrix and its values. That's where Hypothesis could come into play.

tests/unit/test_output_fcidump_parser.py
```diff
@@ -144,3 +145,111 @@ def test_eri_tensor_shape_and_symmetry() -> None:
...
+
+
+@pytest.mark.unit
+@pytest.mark.output
+def test_hcore_matrix_zero_index_raises() -> None:
+    """An orbital index of 0 (1-based convention) must raise instead of wrapping around."""
```
timmyte: What would happen for negative indices?
This could also be covered here.

timmyte (changes requested):
Thanks for introducing this.
Please have a look again especially at the unit tests.

src/opi/output/fcidump.py
haneug: Good point, I perform these checks anyway!

tests/unit/test_output_fcidump_parser.py
haneug: Anything smaller than 1 will raise an error here. I added a test for the negative values as well.

tests/unit/test_output_fcidump_parser.py
haneug: That is a good point. I am now using hypothesis to test different sizes ranging form 2x2 to 8x8. Also, I added a small test for what value is actually written (the last one).

src/opi/output/fcidump.py
```diff
@@ -201,7 +202,7 @@ def from_arrays(
             norb=norb,
             nelec=nelec,
             ms2=ms2,
-            orbsym=list(orbsym) if orbsym is not None else [1] * norb,
+            orbsym=list(orbsym) if orbsym else [1] * norb,
```
timmyte, line 232: Please add a brief comment, why the fall back is an array with only one's.

tests/unit/test_output_fcidump_parser.py
```diff
@@ -199,14 +201,54 @@ def test_from_arrays_asymmetric_raises() -> None:
...
+
+
+@pytest.mark.parametrize(
+    "one_electron, expected",
+    [
+        ({(1, 2): 0.1, (2, 1): 0.2}, 0.2),
+        ({(2, 1): 0.2, (1, 2): 0.1}, 0.1),
+    ],
+)
+def test_hcore_matrix_last_write_wins(one_electron, expected) -> None:
```
timmyte: Could please clarify what this test is for.
Also, what you mean by "last write wins". I would describe such phrases in the docstring (btw: which seems to missing :D).

tests/unit/test_output_fcidump_parser.py
```diff
@@ -219,6 +261,16 @@ def test_hcore_matrix_zero_index_raises() -> None:
         dump.hcore_matrix
 
 
+@pytest.mark.unit
+@pytest.mark.output
+def test_hcore_matrix_negative_index_raises() -> None:
+    """An orbital index that is negative must raise instead of wrapping around."""
```
timmyte: Raise what?
Please name the exception in the docstring.

timmyte (changes requested):

tests/unit/test_output_fcidump_parser.py
haneug: I added a docstring and chose a more descriptive name for the function.

timmyte (approved):

PR #269, opened 2026-07-24
fix: more precise unit conversion factors

haneug:
## Closes Issues
Closes None

## Description
- To be consistent with ORCA the unit conversion constants need to be updated.

---

### Changed 

- Updated unit conversion constants to be consistent with ORCA (#269).


timmyte (approved):
Infinite precision :D

PR #273, opened 2026-07-30
feat: add Structure.to_ase() and adapt Structure.from_ase()

### LINKED ISSUE #271: Add to_ase() to the Structure class
Better integration with the ase is desirable. As simple next step is to add a `to_ase()` method to the Structure class.

haneug:
## Closes Issues
Closes #271 

## Description
- Add a new function `Structrue.to_ase()` to generate an ase atoms object from the OPI structure. Since this requires importing the ase I added it as an optional dependency. The implementation comes with a new example test for to_ase() which also runs the complete roundtrip to and from ase. Since in the bare OPI Structure class we do not have atomic charges or spins the are currently not considred in `to_ase()`, instead the total charge and multiplicity are written to `atoms.info` as `charge` and `spin`. To make the roundtrip work properly I also updated `from_ase` to use these `atoms.info` fields if not per-atom data is available for the charges and spins. 

---

### Added 

- Added `to_ase()` to convert a `Structure` into an ASE `Atoms` object (#273)
- `Structure.from_ase()` now falls back to the `charge` and `spin` entries of `Atoms.info` if ASE's per-atom `initial_charges` / `initial_magnetic_moments` arrays are unset (#273)


pyproject.toml
```diff
@@ -38,6 +38,9 @@ Issues = "https://github.com/faccts/opi/issues"
 Changelog = "https://github.com/faccts/opi/blob/main/CHANGELOG.md"
 
 [project.optional-dependencies]
+ase = [
+    "ase~=3.23",
```
timmyte: This is quite a strict version constraint. And `3.23` is also quite dated. Any particular reason, why it has to be this strict?
Can we go with `ase<4,>=3.23`?

pyproject.toml
```diff
@@ -68,6 +71,7 @@ dev = [
     "nox>=2025.5.1",
 ]
 tests = [
+    "ase~=3.23",
```
timmyte: See above.

src/opi/input/structures/structure.py
```diff
@@ -1051,20 +1061,83 @@ def from_ase(
...
+        ------
+        ImportError
+            If ASE is not installed.
+        ValueError
+            If this structure contains no real atoms.
+        """
+        # > ASE is an optional dependency and must not be required to import OPI,
+        # > hence the in-method import. The module-level `AseAtoms` name is a
+        # > TYPE_CHECKING-only alias that exists solely for annotations.
+        try:
```
timmyte: I would suggest to have a look how @crjacinto did this with `molbar` and follow his pattern. It's more elegant.

src/opi/input/structures/structure.py
```diff
@@ -1051,20 +1061,83 @@ def from_ase(
...
+            raise ImportError("ASE is not installed. Install it with: pip install ase") from err
+
+        if not self.real_atoms:
+            raise ValueError(
+                f"{self.__class__.__name__}: structure contains no real atoms; "
+                "cannot build ASE Atoms object."
+            )
+
+        # > Only real Atom instances are converted
+        real_indices = [i for i, a in enumerate(self.atoms) if type(a) is Atom]
```
timmyte: We should a function to determine if an `atom` is real atom. I'm think to recall to @crjacinto added this in one of his PRs.
Anyhow, if not then it would be great if you could add such a function, as it will be a repeating pattern to sort out non-real atoms.

timmyte (changes requested):
Great that you extended the interface to ASE. This will be a nice addition.

src/opi/input/structures/structure.py
haneug: We have the property `real_atoms` but this returns a list of the real Atoms and the index information is lost. The real index pattern is used ~4 times in the `Structure` class. I now added an additional property `real_atom_indices` which just returns `list[int]` with the positions.

src/opi/input/structures/structure.py
haneug: I tried to mimic the molbar approach to importing.

src/opi/input/structures/structure.py
```diff
@@ -116,6 +117,13 @@ def real_atoms(self) -> list[Atom]:
         """
         return [a for a in self.atoms if type(a) is Atom]
 
+    @property
+    def real_atom_indices(self) -> list[int]:
+        """
+        Return the indices of real `Atom` instances as provided by `real_atoms()`.
```
timmyte, line 123:
```suggestion
        Return the indices of real `Atom` instances as provided by `real_atoms`.
```

src/opi/utils/ase.py
```diff
@@ -0,0 +1,130 @@
...
+    @wraps(func)
+    def wrapper(*args: Any, **kwargs: Any) -> _T:
+        if not _ase_available():
+            raise ImportError("ASE is not installed. It is an optional dependency of OPI.")
+        return func(*args, **kwargs)
+
+    return wrapper
+
+
+def _import_ase(func: Callable[..., _T]) -> Callable[..., _T]:
```
timmyte, line 61: Thanks for adding it like this.
I just realize that we don't need three routines for that. We just need this one here. I would rename it to `requires_ase()` and every function in every module that requires ASE is just decorated with that.

It does everything we need. It checks, if has been imported and if not tries to import it.

I don't see a benefit with the current `requires_ase()`. Having one decorator makes the project better scalable.

src/opi/utils/ase.py
```diff
@@ -0,0 +1,130 @@
...
+    """
+    Decorator that lazily imports ASE on first call and caches it.
+    """
+
+    @wraps(func)
+    def wrapper(*args: Any, **kwargs: Any) -> _T:
+        if globals()["ase_atoms_cls"] is None:
+            module = import_module("ase")
+            globals()["ase_atoms_cls"] = module.Atoms
```
timmyte, line 70: I would not import the module under an alias. I would import it as `Atoms`. If you don't like the like the alias, I would at least preserve the correct casing, i.e. `AseAtoms`.

src/opi/utils/molbar.py
```diff
@@ -82,7 +80,7 @@ def requires_molbar(func: Callable[..., _T]) -> Callable[..., _T]:
     @wraps(func)
     def wrapper(*args: Any, **kwargs: Any) -> _T:
         if not _molbar_available():
-            raise ImportError("MolBar is not installed. Install it with: pip install molbar")
+            raise ImportError("MolBar is not installed. It is an optional dependency of OPI.")
```
timmyte, line 83: Good catch.

tests/examples/test_exmp057_to_ase.py
```diff
@@ -0,0 +1,25 @@
+# ruff: noqa: E402
+import pytest
+
+# > example57 requires ase, which is not installed by default
+# > skip test if ase is not available
+pytest.importorskip("ase", reason="requires ase")
```
timmyte, line 6: Very good!

timmyte (changes requested):
Looks good. I would just try to reduce number of decorators for the ASE import.

src/opi/input/structures/structure.py
```diff
@@ -243,8 +251,7 @@ def centered_structure(self) -> "Structure":
             New `Structure` with centered coordinates.
         """
         new_structure = copy.deepcopy(self)
-        real_indices = [i for i, a in enumerate(new_structure.atoms) if type(a) is Atom]
-        centroid = new_structure.get_coordinates(only_atoms=real_indices).mean(axis=0)
+        centroid = new_structure.get_coordinates(only_atoms=self.real_atom_indices).mean(axis=0)
```
timmyte, line 254: Very smooth!

src/opi/utils/ase.py
```diff
@@ -0,0 +1,51 @@
...
+
+
+# ============================================================
+# Decorators
+# ============================================================
+
+_T = TypeVar("_T")
+
+
+def requires_ase(func: Callable[..., _T]) -> Callable[..., _T]:
```
timmyte, line 23: Very good!!

timmyte (approved):
Many thanks for implementing this.
Especially the decorator you developed for the optional import of ASE. That's very helpful!

PR #274, opened 2026-07-31
feat:add block merge logic

### LINKED ISSUE #165: Implement common arithmetic operations for Blocks
`Blocks` are similar to `dict`, so we should also define common operation between blocks such as addition, subtraction etc.

nakul680:
## Closes Issues
Closes #165 

## Description
- add logic to merge two blocks of the same type. 
- implement __or__ operator for Blocks, right-side gets preference

---

### Added 

- add merge logic for two `Block` instants of same type.(#274)

### Changed 

- change `Input.add_blocks()` to merge two block instants of same type by default, instead of being a no-op.(#274)


src/opi/input/core.py
```diff
@@ -569,10 +570,14 @@ def add_blocks(
...
-                self._blocks[type(block)] = block
-            elif strict:
-                raise ValueError(f"Strict: Block for {block.name} has already been added")
+            else:
+                if strict:
+                    raise ValueError(f"Strict: Block for {block.name} has already been added")
+                elif overwrite:
+                    self._blocks[type(block)] = block
+                else:
+                    existing_block = next(iter(self.get_blocks(type(block)).values()))
```
haneug, line 579: You know the type here so you could get rid of `next(iter(....`

tests/unit/test_input_blocks.py
```diff
@@ -192,3 +193,169 @@ def test_clear_blocks_strict(empty_calc: Calculator):
...
+    right = BlockScf(maxiter=200)
+
+    merged = left | right
+
+    assert merged.maxiter == 200
+
+
+@pytest.mark.unit
+@pytest.mark.input
+def test_block_or_preserves_self_field_left_unset_by_other():
```
haneug, line 230: This is the same test as `test_block_or_combines_disjoint_fields()` only differing by one assert statement. So maybe just merge the tests?

tests/unit/test_input_blocks.py
```diff
@@ -192,3 +193,169 @@ def test_clear_blocks_strict(empty_calc: Calculator):
...
+
+    merged = left | right
+
+    assert merged.maxiter == 100
+    assert left.maxiter == 100
+
+
+@pytest.mark.unit
+@pytest.mark.input
+def test_block_or_returns_new_instance_without_mutating_operands():
```
haneug, line 244: Here the setup is also the same but there are quite a lot of different assert statements. Maybe you can put the setup in a fixture?

tests/unit/test_input_blocks.py
```diff
@@ -192,3 +193,169 @@ def test_clear_blocks_strict(empty_calc: Calculator):
...
+
+    assert merged.nuclei == right.nuclei
+    assert merged.gtensor == 1  # preserved from `left`, untouched by `right`
+
+
+@pytest.mark.unit
+@pytest.mark.input
+def test_block_or_arbitrary_options_are_not_preserved():
+    """Test for `Block.__or__()`: arbitrary options added via `add_option()` live on a private,
+    non-model attribute and are dropped by the merge (`model_dump()` does not include them), even
```
haneug, line 280: This should definitely be documented in the docstring of `__or__`. Would it be possible to also merge the options?

tests/unit/test_input_blocks.py
```diff
@@ -192,3 +193,169 @@ def test_clear_blocks_strict(empty_calc: Calculator):
...
+    merged = left | right
+
+    assert not merged.has_option("foo")
+
+
+@pytest.mark.unit
+@pytest.mark.input
+def test_block_or_other_default_overrides_self_explicit_value():
+    """Test for `Block.__or__()`: `aftercoord` has a non-`None` default (`False`), so it is always
+    included by `model_dump(exclude_none=True)`. As a result, `other`'s default silently overrides
```
haneug, line 295: This should also be documented with `__or__`

haneug (changes requested):

src/opi/input/blocks/base.py
```diff
@@ -169,6 +169,25 @@ def format_orca(self) -> str:
 
         return s
 
+    def __or__(self, other: "Block") -> "Block":
+        """
+        Merges two instances of `Block`. If a common attribute exists in both `self` and `other`, the value in `other` will be given precedence.
+        Parameters
```
timmyte, line 175:
```suggestion
        Merges two instances of `Block`. If a common attribute exists in both `self` and `other`, the value in `other` will be given precedence.
        
        Parameters
```

src/opi/input/core.py
timmyte: Also did you consider that `next()` raises `StopIteration`, when the iterator is empty.
Regardless how you finally implement this, make sure to also cover this possibility.
You can try yourself:
```
next(iter({}.values()))
```

tests/unit/test_input_blocks.py
```diff
@@ -273,3 +282,274 @@ def test_clear_blocks_strict(empty_calc: Calculator):
...
+
+    merged = left | right
+
+    assert merged.get_option("first") == "1"
+    assert merged.get_option("second") == "2"
+
+
+@pytest.mark.unit
+@pytest.mark.input
+def test_arbitrary_block_or_result_is_writable_to_orca_input():
```
haneug, line 474: Has the same setup as the previous test. Would you consider merging them?

haneug (approved):

tests/unit/test_input_blocks.py
nakul680: Im not sure there is any reason to merge these tests , the setup is relatively simple so i don think its worth creating a fixture or something from them

PR #276, opened 2026-08-03
feat: add Block to create arbitrary blocks

### LINKED ISSUE #275: Add functionality to create arbitrary blocks
Currently it is impossible to to create an arbitrary blocks in OPI, so the workaround right now is to simply add an arbitrary string.

nakul680:
## Closes Issues
Closes #275 

## Description
- add `Block` which will allow users to create arbitrary blocks. Any options in an instance of `Block` is stored as an arbitrary option. 
- upon initialization of an `Block` instance, user will be able to work with it the same way they would for other blocks implemented in OPI, except instead of using their type to fetch a block or remove, the user would have to use the assigned name or an instant.
- Users can now fetch blocks or remove blocks using the name of the block
- Blocks are now stored in a dict keyed by their names instead of the type they are stored as.

---

### Added 

- Add `Block` to allow for creation of arbitrary blocks. (#276)
- Add functionality to fetch, search or remove a block using the ORCA name of the block. (#276)

### Changed 
- `Input` now stores blocks under the name of the ORCA block they model instead of under their `BlockABC` class. Hence `Input.blocks` and the dictionary returned by `Input.get_blocks()` are keyed by that name, e.g. `calc.input.get_blocks(BlockScf)["scf"]`. (#276)

### Fixed
- The block methods of `Input` now raise a `TypeError` when passed something that is neither a block, a block class nor a block name, instead of failing with an `AttributeError`. `Input.add_blocks()` also rejects block classes, as only an instance carries the options of a block (#276).
- `BlockABC` can no longer be instantiated, as it is abstract and models no ORCA block. `BlockABC.name` now reports that a block defines no ORCA block name instead of failing on the missing private attribute `_name` (#276).


src/opi/input/blocks/arbitrary_block.py
```diff
@@ -0,0 +1,74 @@
...
+
+__all__ = [
+    "ORCABlock",
+]
+
+
+_NAME_PATTERN = re.compile(r"^[a-zA-Z_][a-zA-Z_0-9]*$")
+
+
+class ORCABlock(Block):
```
timmyte, line 14: 1) If a module only contains one distinct class, make sure that both have the same name (except for casing conventions). 
```
from opi.input.blocks.arbitrary_block import ArbitraryBlock
# is more intuitive than
from opi.input.blocks.arbitrary_block import ORCABlock
```

2) IMO the name does not reflect very well the purpose of this class.
* I would rather rename `Block` to `BlockABC` (or similar) and name this class `Block`
* Or name this one here `CustomBlock`, `ModelBlock`, etc.

@haneug 
What's your take on the second point?

src/opi/input/blocks/arbitrary_block.py
```diff
@@ -0,0 +1,74 @@
...
+        name = self.normalize(name)
+        self._name = name
+        if values:
+            self._arbitrary = NoCaseDict(values)
+
+    def normalize(self, name: str) -> str:
+        """
+        Normalize the arbitrary block name. Strips leading '%' and whitespaces.
+        Parameters
+        ----------
```
timmyte, line 49:
```suggestion
        Normalize the arbitrary block name. Strips leading '%' and whitespaces.
        
        Parameters
        ----------
```

src/opi/input/blocks/arbitrary_block.py
```diff
@@ -0,0 +1,74 @@
...
+        -------
+        str
+            Normalized string.
+
+        Raises
+        ------
+        ValueError
+            If user-given name is invalid, or an ORCa block with that name is already implemented.
+        """
+        name = name.strip("% ").lower()
```
timmyte, line 62:
```suggestion
        name = name.strip().removeprefix("%").lower()
```

src/opi/input/blocks/arbitrary_block.py
```diff
@@ -0,0 +1,74 @@
...
+
+        if not _NAME_PATTERN.fullmatch(name):
+            raise ValueError("Invalid name for ORCA block")
+
+        block_class = Block.get_block_class(name)
+
+        if block_class:
```
timmyte, line 69: Very good!

src/opi/input/blocks/base.py
```diff
@@ -26,13 +26,74 @@ class Block(BaseModel, ABC):
...
+    # > Registry of the subclasses of `Block`.
+    # > Being a `ClassVar`, it is a single object shared by the whole class hierarchy.
+    # > It is filled by `__pydantic_init_subclass__()`, hence it only contains the
+    # > subclasses whose defining module has been imported.
+    _registry: ClassVar[dict[str, type["Block"]]] = {}
+
+    @classmethod
+    def __pydantic_init_subclass__(cls, **kwargs: Any) -> None:
+        """
+        Register a subclass of `Block` in `Block._registry` under its block name.
```
timmyte, line 50: Please add what we need the registry for.

src/opi/input/blocks/base.py
```diff
@@ -26,13 +26,74 @@ class Block(BaseModel, ABC):
...
+        super().__pydantic_init_subclass__(**kwargs)
+        name = cls.get_block_name()
+        # > Subclasses without a class-level `_name` are named at runtime and cannot be registered.
+        if name is not None:
+            # > The first registration wins, so a subclass cannot displace the block class it derives from.
+            Block._registry.setdefault(name, cls)
+
+    @classmethod
+    def get_block_name(cls) -> str | None:
+        """
```
timmyte, line 67: I would go with an empty string as null value.

src/opi/input/core.py
```diff
@@ -536,6 +536,48 @@ def clear_simple_keywords(self, *, strict: bool = False) -> None:
     # ----------------------------------------------------------------------
     # > BLOCKS
     # ----------------------------------------------------------------------
+    @staticmethod
+    def _block_key(block: Block | type[Block] | str, /) -> type[Block] | str:
```
timmyte, line 540: I thought we discussed to abandon `type(Block)` as identifier and stick to the names.
We can keep this as valid input parameter from the function to be more flexible, but I would always return just the name.
Is there a reason why you kept it?

Also, I would avoid functions whose return data type depends on the input parameter's.

src/opi/input/core.py
```diff
@@ -598,49 +642,43 @@ def remove_blocks(self, *blocks: Block, strict: bool = False) -> None:
                 return
 
         for block in blocks:
-            t_block = type(block)
-            if t_block in self._blocks:
-                self._blocks.pop(t_block)
+            key = self._block_key(block)
```
timmyte, line 645: Very good to have a function that resolves the identifier!!

src/opi/input/core.py
```diff
@@ -536,6 +536,48 @@ def clear_simple_keywords(self, *, strict: bool = False) -> None:
...
+        Raises
+        ------
+        ValueError
+            If `block` is a class that is named at runtime, as such a class does not identify
+            a single block.
+        """
+        if isinstance(block, str):
+            name = block
+        elif isinstance(block, type):
+            block_name = block.get_block_name()
```
timmyte, line 568: Every class in Python is an instance of `type`, so `block.get_block_name()` can throw an `AttributeError`. Take that into account!
It should be documented or handled properly, but we don't want OPI crashing because of that.

src/opi/input/blocks/arbitrary_block.py
```diff
@@ -0,0 +1,74 @@
...
+    def __init__(self, name: str, values: dict[str, str] | None = None, aftercoord: bool = False):
+        """
+
+        Parameters
+        ----------
+        name: Name of arbitrary block
+
+        values: Values to be added to the block. They will be added as arbitrary options, so both key and value must be strings.
+
+        aftercoord: Sets whether block will appear before or after coordinates in the ORCA .inp file.
```
timmyte, line 36:
```suggestion
        name: 
            Name of arbitrary block
        values:
            Values to be added to the block. They will be added as arbitrary options, so both key and value must be strings.
        aftercoord:
            Sets whether block will appear before or after coordinates in the ORCA .inp file.
```

tests/unit/test_input_arbitrary_block.py
```diff
@@ -0,0 +1,141 @@
...
+def test_arbitrary_block_implemented_name(name: str):
+    """Test that an `ORCABlock` cannot take the name of an implemented block."""
+    with pytest.raises(ValueError):
+        ORCABlock(name)
+
+
+@pytest.mark.unit
+@pytest.mark.input
+def test_arbitrary_block_registry():
+    """Test that implemented blocks are registered by name, while `ORCABlock` is not."""
```
timmyte, line 87: I don't quite understand why we need that exception for `ORCABlock`. Could please explain.

timmyte (changes requested):

src/opi/input/blocks/arbitrary_block.py
haneug: I agree and would prefer to go with `BlockABC` for the ABC

examples/exmp024_blocks/job.py
```diff
@@ -49,9 +49,9 @@ def run_exmp024(
 
     # add block, get it, modify it, and add it back to the Calculator
     calc.input.add_blocks(BlockMethod(d3s6=0.64, d3a1=0.3065, d3s8=0.9147, d3a2=5.0570))
-    block = calc.input.get_blocks(BlockMethod)
-    # modify the %method block
-    block[BlockMethod].d3s6 = 0.4
+    method_block = calc.input.get_block(BlockMethod)
+    # modify the %method block, which is keyed by its ORCA block name
```
timmyte, line 53: I think the comment is misleading.
You say it's keyed by *foo* but use *bar* to get the block.

src/opi/input/blocks/base.py
```diff
@@ -26,13 +31,120 @@ class Block(BaseModel, ABC):
...
+            Fields of the block, passed on to pydantic.
+
+        Raises
+        ------
+        TypeError
+            If `BlockABC` itself is instantiated.
+        """
+        # > `ABC` alone does not prevent instantiation, as `BlockABC` defines no abstract method.
+        # > Without this guard, a bare `BlockABC` would be a block without a name, and every
+        # > access to `BlockABC.name` would fail with an `AttributeError`.
```
timmyte, line 68: Very good!

src/opi/input/core.py
timmyte: I think the `isinstance()` is obsolete when you use `issubclass()`.

timmyte (approved):
Good to go from my side.
Please just regard my two remarks before merging.

src/opi/input/core.py
nakul680: It is needed for the second condition, so without the isinstance() the add_blocks() would be broken too

PR #279, opened 2026-08-12
`_orca_environment()` mutate environ dict without breaking references

### LINKED ISSUE #264: orca-pi breaks expected environment behaviour
**Describe the bug**
orca-pi breaks the expected `os.environ` behaviour be reassigning it to a copy. This breaks any tool that stores a reference to the `os.environ` object (monkeypatch, but also `os.getenv` if it's ever memoized, etc.).

**To Reproduce**
```
"""
Reproduces the opi._orca_environment bug where os.environ = dict_copy
breaks pytest monkeypatch (and any code holding a reference to os.environ).

Does NOT require ORCA to be installed.
"""
import os
import pytest


def _orca_environment_buggy(fn):
    """Minimal reproduction of opi.execution.core._orca_environment."""
    def wrapper(*args, **kwargs):
        org_env = os.environ.copy()   # plain dict copy
        try:
            os.environ["PATH"] = "/fake/orca/bin:" + os.environ.get("PATH", "")
            return fn(*args, **kwargs)
        finally:
            os.environ = org_env      # replaces the module attribute with a new dict
    return wrapper


@_orca_environment_buggy
def simulate_orca_run():
    pass  # pretend ORCA ran here


def test_env_leak(monkeypatch):
    simulate_orca_run()                    # os.environ is now dict D1 (a copy)
    monkeypatch.setenv("ORCA_MEMORY", "1") # monkeypatch captures D1
    simulate_orca_run()                    # os.environ is now dict D2 (copy of D1)
    # monkeypatch teardown will del D1["ORCA_MEMORY"]
    # but os.environ == D2, which still has ORCA_MEMORY=1


def test_env_still_set():
    """Passes when run alone, fails when run after test_env_leak."""
    assert os.environ.get("ORCA_MEMORY") is None, (
        f"ORCA_MEMORY leaked from previous test: {os.environ.get('ORCA_MEMORY')!r}\n"
        f"os.environ type: {type(os.environ).__name__}  "
        f"(expected os._Environ, got plain dict — set by _orca_environment)"
    )
```

**Expected behavior**
monkeypatch teardown should work

**Version:**
 - opi 2.0

**Suggested fix pattern (Claude)**
```
# current (broken for monkeypatch)
org_env = os.environ.copy()
try:
    add_to_env("PATH", ...)
    ...
finally:
    os.environ = org_env

# fixed
saved_path = os.environ.get("PATH")
saved_ldlib = os.environ.get("LD_LIBRARY_PATH")
try:
    add_to_env("PATH", ...)
    ...
finally:
    if saved_path is not None: os.environ["PATH"] = saved_path
    if saved_ldlib is not None: os.environ["LD_LIBRARY_PATH"] = saved_ldlib
```

timmyte:
## Closes Issues
Closes #264
...

## Description
- `_orca_environment()` now clears and updates the dict in-place. This will preserve any references to the dict.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -43,7 +43,7 @@
 - Updated deprecated `typing` types to be compliant with Python >=3.11 guidelines (#216)
 - Fixed a typo in the badge for the OPI paper (#222).
 - Fixed `Structure.nelectrons` for structures containing ghost atoms (#268).
-- Fixed `_orca_environment()` which now makes changes to `os.environ` in-place without breaking any reference to that dict.
+- Fixed `_orca_environment()` which now makes changes to `os.environ` in-place without breaking any reference to that dict (#279).
 
 ## [2.0.0] - 2026-02-10
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,7 +21,7 @@
 - Added `get_frequencies`, `get_imaginary_frequencies`, `is_pes_minimum`, and `is_pes_transition_state` to the `Output` class (#247).
 - Add `functions to clean up files created by ORCA jobs.(#262)
 - `Output.parse()`, `Output.parse_property()`, `Output.parse_gbw()`, and `Output.__init__()` now accept a `strict` parameter (default True). When set to False, output fields that fail Pydantic validation are silently set to None and a UserWarning is emitted instead of raising a ValidationError (#248).
-- Added `rmsd` and `rmsd_kabsch` to calculate RMSD without and with rotational alignment to `Structure` class (#230). 
+- Added `rmsd` and `rmsd_kabsch` to calculate RMSD without and with rotational alignment to `Structure` class (#230).
 - Added `calc_rotational_constants` to calculate molecular rotational constants to `Structure` class (#230).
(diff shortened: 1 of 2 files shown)


timmyte:
@kjh-giesbertz 
Would this solution also work for you?

CHANGELOG.md
```diff
@@ -43,16 +43,17 @@
 - Updated deprecated `typing` types to be compliant with Python >=3.11 guidelines (#216)
 - Fixed a typo in the badge for the OPI paper (#222).
 - Fixed `Structure.nelectrons` for structures containing ghost atoms (#268).
+- Fixed `_orca_environment()` which now makes changes to `os.environ` in-place without breaking any reference to that dict.
```
nakul680: Missing PR number

```suggestion
- Fixed `_orca_environment()` which now makes changes to `os.environ` in-place without breaking any reference to that dict (#279).
```

nakul680 (approved):

kjh-giesbertz:
> @kjh-giesbertz Would this solution also work for you?

I have run the purest snippet and it indeed seems to fix it. Thanks!

PR #280, opened 2026-08-17
feat(GbwResults): add `from_file()` and `get_structure()`

haneug:
## Closes Issues
Closes None

## Description
- Adds `get_structure()` and `from_file()` to `GbwResults`. This makes it easier to improve a single gbw file and it allows simple access to the structure when no property file is present. The gbw structure is also exposed via `Output.get_structure_from_gbw()`

---

### Added 

- Added `GbwResults.get_structure()`, `GbwResults.from_gbw_file()`, and `Output.get_structure_from_gbw()` to obtain a `Structure` from a gbw file or gbw JSON (#280).

### DIFF
--- a/src/opi/output/core.py
+++ b/src/opi/output/core.py
@@ -1503,6 +1503,36 @@ class Output:
             structure.multiplicity = mult
         return structure
 
+    def get_structure_from_gbw(self, *, index: int = 0) -> Structure | None:
+        """
+        Returns the structure stored in a gbw file as `Structure` object.
+        Silently returns None if no structure is available.
+
+        In contrast to `get_structure()`, which reads the geometry from the property JSON, this
+        getter uses the gbw JSON. It therefore offers no fragment IDs and no `.out` file fallback,
+        but it does distinguish ghost atoms from real ones.
+
+        Parameters
+        ----------
+        index : int, default: 0
+            Index (>= 0) of the gbw file in `self.results_gbw`. The default 0 refers to the main
+            gbw file.
+
+        Returns
+        ----------
+        structure: Structure | None
+            Structure generated from the gbw data or None if no structure is available.
+        """
+        results_gbw = self._safe_get("results_gbw", index)
+        if results_gbw is None:
+            return None
+
+        try:
+            return cast(GbwResults, results_gbw).get_structure()
+        except ValueError:
+            # > `GbwResults.get_structure()` raises on malformed atom data
+            return None
+
     def _get_cartesians(
         self, index: int, /
     ) -> list[tuple[StrictStr, StrictFiniteFloat, StrictFiniteFloat, StrictFiniteFloat]] | None:
--- a/src/opi/output/models/json/gbw/gbw_results.py
+++ b/src/opi/output/models/json/gbw/gbw_results.py
@@ -1,9 +1,17 @@
+from pathlib import Path
+
 from pydantic import Field
 
+from opi.execution.core import Runner
+from opi.input.structures.atom import Atom, GhostAtom
+from opi.input.structures.coordinates import Coordinates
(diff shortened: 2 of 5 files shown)


src/opi/output/models/json/gbw/gbw_results.py
```diff
@@ -26,3 +34,120 @@ class GbwResults(JSONLoadable):
...
+        if not gbw_file.is_file():
+            raise FileNotFoundError(f"File {gbw_file} not found")
+
+        # > `orca_2json` maps every binary suffix onto the same "<basename>.json". Reusing an
+        # > existing JSON is therefore only safe for ".gbw" itself, otherwise we could pick up the
+        # > JSON belonging to a different binary (e.g. "job.gbw" when asked for "job.loc").
+        force = force or gbw_file.suffix != GbwSuffix.GBW.value
+
+        runner = Runner(working_dir=gbw_file.parent)
+        runner.create_gbw_json(gbw_file.stem, force=force, config=config, suffix=gbw_file.suffix)
```
nakul680, line 78: after looking through `Runner.create_gbw_json()` it seems like if the gbw json already exists and `force=False` , the `config` that the user sets will be dropped

src/opi/output/models/json/gbw/gbw_results.py
```diff
@@ -26,3 +34,120 @@ class GbwResults(JSONLoadable):
...
+            Raised if the created JSON is invalid.
+        """
+        gbw_file = Path(gbw_file).expanduser().resolve()
+        if not gbw_file.is_file():
+            raise FileNotFoundError(f"File {gbw_file} not found")
+
+        # > `orca_2json` maps every binary suffix onto the same "<basename>.json". Reusing an
+        # > existing JSON is therefore only safe for ".gbw" itself, otherwise we could pick up the
+        # > JSON belonging to a different binary (e.g. "job.gbw" when asked for "job.loc").
+        force = force or gbw_file.suffix != GbwSuffix.GBW.value
```
nakul680: In my opinion if we want use the JSON that exists already from the same binary it would be better to overwrite every time, unless `orca2json` is an expensive operation. So maybe `strict` should be the default option with the reuse being the argument that the user can set, i.e a `reuse_json` argument?

nakul680: The comparison of the gbw suffix is done on the raw value , we could instead use the `_missing_()` that exists in GbwSuffx to handle case sensitivity and leading dots

src/opi/output/core.py
```diff
@@ -1503,6 +1503,36 @@ def get_structure(
             structure.multiplicity = mult
         return structure
 
+    def get_structure_from_gbw(self, *, index: int = 0) -> Structure | None:
```
nakul680, line 1506: is it possible for `Output.get_structure_from_gbw()` to yield a different `Structure` object than `Output.get_structure`?

haneug: In principle they should be the same structure, however these are different sources of the same information and they could diverge e.g. by just replaving the gbw file. There, I would like to keep them separated. The function `get_structure()` should still be the preferred function.

src/opi/output/models/json/gbw/gbw_results.py
haneug: I removed this check since we force now by default

haneug: I agree and introduced the reuse_json argument

src/opi/output/models/json/gbw/gbw_results.py
haneug: I added a warning about this in the docstring.

nakul680 (approved):

PR #283, opened 2026-08-20
feat: Update to get_energies()

### LINKED ISSUE #281: Missing components in `Output.get_energies()`
The function `Output.get_energies()` does not fetch all the contributions to get to the final energy. I think ideally if all components are added up, one should arrive to the result of `Output.get_final_energy()` missing components I have encountered so far:

- vdw contribution
- gCP correction
- gCP+bas correction (present in e.g. a hf-3c calculation)
- SRB correction (present in e.g. a b97-3c calculation)

I think these components should also be present in the dict returned by `Output.get_energies()`

haneug:
## Closes Issues
Closes #281

## Description
- added gcp and vdw contribution to get energies
- added property `energy` to the `Energy` base class for easier access
- added an enum to keep track of the energy types of get_energies

---

## Release Notes

### Added 

- `Output.get_energies()` now also reports the corrections that ORCA adds on top of the total energy of the electronic structure method: the dispersion correction (`VdW`) and the geometrical counterpoise correction (`gCP`) (#283).
- Added `Output.get_vdw_correction()`, `Output.get_gcp_correction()`, `Output.get_rocis_energies()` and `Output.get_cipsi_energies()` to access the dispersion correction, the gCP correction and the energies of ROCIS and ICE-CI/CIPSI calculations (#283).
- Added the missing field `numofroots` to `RoCisEnergy` (#283).
- Added the `Energy.energy` property as well as `reference_energy` and `correlation_energy` on the correlated energies, which return the energies as plain floats instead of the nested list (#283).
- Added `EnergyType`, which enumerates and documents the known keys of the dictionary returned by `Output.get_energies()` (#283).
 
### Fixed 

- The docstrings of `Mp2EnergyBase` and `AutoCiEnergy` documented `refenergy` and `correnergy` as plain floats, while ORCA writes them as a nested list holding a single element (#283).


src/opi/output/core.py
```diff
@@ -1361,14 +1367,27 @@ def get_energies(self, *, index: int = -1) -> dict[str, Energy] | None:
...
+
+        The corrections are wrapped into `Energy` objects carrying only the method and the energy
+        itself, so their value is accessed the same way as for the other entries, i.e. via the
+        `Energy.energy` property. Adding them to the total energy of the highest-level method
+        present yields `Output.get_final_energy()`. Everything else that ORCA reports about them is
+        available from `Output.get_vdw_correction()` and `Output.get_gcp_correction()`.
+
+        Not every energy that ORCA can report shows up here: methods whose results are not modelled
+        as an `Energy` have their own getters, namely `Output.get_rocis_energies()` and
+        `Output.get_cipsi_energies()`.
```
nakul680: Just a doubt i have here, is the user still able to use the original strings to get the energy instead of the enum member?

tests/examples/test_exmp002_scf_ccsdt.py
```diff
@@ -19,4 +20,4 @@ def test_exmp002_scf(example_input_file, tmp_path) -> None:
     # Assert negative final energy
     assert output.get_final_energy() < 0
     # Assert correlation energy
-    assert output.get_energies()["MDCI(SD(T))"].correnergy[0][0] < 0
+    assert output.get_energies()[EnergyType.MDCI_SD_T].correlation_energy < 0
```
nakul680: assuming one could also use plain strings as a key for get_energies() , I'm not sure this version of the API is what we want to use in examples, since it could give the user the impression of a more complicated API. This however is also subjective so I think maybe we should discuss it in the OPI JF further.

haneug:
I will reopen this in a separate PR

PR #284, opened 2026-08-20
fix: clamp negative ORCA timings

### LINKED ISSUE #263: Handle negative timings
**Describe the bug**
For very short calculations timings can come out negative causing due to floating point errors. This causes a validation error getting the property results

**To Reproduce**
This was for a ccsd calculation on H2O in def2-svp, but it flaky, so not well reproducible and very system dependent.

**Trace back**
```
        # // Parse JSON
        self.property_json_data = self._process_json_file(self.property_json_file)
        # > Check in property JSON whether version fits:
        if self.do_version_check:
            self.check_version()
>       self.results_properties = PropertyResults(**self.property_json_data)
                                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
E       pydantic_core._pydantic_core.ValidationError: 2 validation errors for PropertyResults
E       calculation_timings.prop
E         Input should be greater than or equal to 0 [type=greater_than_equal, input_value=-0.6242519999999999, input_type=float]
E           For further information visit https://errors.pydantic.dev/2.13/v/greater_than_equal
E       calculation_timings.sum
E         Input should be greater than or equal to 0 [type=greater_than_equal, input_value=-0.1279849999999999, input_type=float]
E           For further information visit https://errors.pydantic.dev/2.13/v/greater_than_equal

/home/giesbertz/miniforge3/envs/livdft-fix/lib/python3.11/site-packages/opi/output/core.py:228: ValidationError
```

**Expected behavior**
I suppose the best solution is to clamp timings to be non-negative. An alternative would be to allow for negative timings, but since they are meaningless (though give an impression how imprecise timings are) I would go for clamping.

**Version:**
 - ORCA 6.1.1
 - OPI 2.0

**Additional context**

haneug:
## Closes Issues
Closes #263

## Description
- Clamp negative timings. Also adds a getter `Output.get_timings`.

---

### Added 

- Added `Output.get_timings()` to access the timings of the calculation steps (#284).
 
### Fixed 

- Negative calculation timings occasionally reported by ORCA are now clamped to zero instead of raising a `ValidationError` (#284).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,7 +37,7 @@
 - `Structure.from_ase()` now falls back to the `charge` and `spin` entries of `Atoms.info` if ASE's per-atom `initial_charges` / `initial_magnetic_moments` arrays are unset (#273)
 - Add `Block` to allow for creation of arbitrary blocks. (#276)
 - Add functionality to fetch, search or remove a block using the ORCA name of the block. (#276)
-- Added `Output.get_timings()` to access the timings of the calculation steps (#XXX).
+- Added `Output.get_timings()` to access the timings of the calculation steps (#284).
 
 ### Changed
 - Refactored methods from Runner into BaseRunner (#193)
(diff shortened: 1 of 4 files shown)


nakul680 (approved):
Looks good!

PR #288, opened 2026-08-26
feat: add input merge logic

### LINKED ISSUE #277: Input merge logic
Add logic to merge two `Input` objects. Preference would be given to the right side object.

nakul680:
## Closes Issues
Closes #277

## Description
- Allows the merging of two `Input` objects.
- Simple keywords will be merged with duplicates dropped, while blocks will be merged according to existing block merge logic.
- Arbitrary strings will follow the same logic as simple keywords, with both lists of arbitrary strings being combined.

---

### Added 

- Add merge logic for two `Input` objects, via `Input.__or__()`: simple keywords are combined without duplicates, blocks that model the same ORCA block are merged, arbitrary strings are concatenated, and `ncores`, `memory` and `moinp` are taken from the right-hand input if set. The right-hand input takes precedence throughout, blocks are copied into the result, and neither operand is modified.(#288)

### Changed 

- `Input.add_arbitrary_string()` now also accepts an `ArbitraryString` instance next to a plain `str`, so that an arbitrary string can be added together with the position it carries. (#288)
- `SimpleKeyword` now compares equal case-insensitively, matching `SimpleKeyword.format_orca()`, which lowercases the keyword, and ORCA itself, which ignores the case. Hence keywords that only differ in case count as duplicates of each other in `Input.add_simple_keywords()`, `Input.remove_simple_keywords()`, `Input.has_simple_keyword()` and `Input.get_simple_keyword()`. (#288)

### DIFF
--- a/src/opi/input/core.py
+++ b/src/opi/input/core.py
@@ -842,23 +842,26 @@ class Input:
     # > ARBITRARY STRINGS
     # ----------------------------------------------------------------------
     def add_arbitrary_string(
-        self, string: str, /, *, pos: ArbitraryStringPos | str | None = None
+        self, string: str | ArbitraryString, /, *, pos: ArbitraryStringPos | str | None = None
     ) -> None:
         """
         Add an arbitrary string.
 
         Parameters
         ----------
-        string : str
+        string : str | ArbitraryString
             Arbitrary string to be added
         pos : ArbitraryStringPos | str | None, default: None
             Coordinates of the arbitrary string within the input file.
             For details look into the `ArbitraryStringPos` class.
         """
+        if isinstance(string, ArbitraryString):
+            arb_string = string
+        else:
+            arb_string = ArbitraryString(string=string)
+            if pos is not None:
+                arb_string.pos = ArbitraryStringPos(pos)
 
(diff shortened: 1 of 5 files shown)


src/opi/input/core.py
```diff
@@ -907,3 +910,72 @@ def clear_arbitrary_strings(self, *, strict: bool = False) -> None:
...
+        """
+        if not isinstance(other, Input):
+            return NotImplemented
+
+        # > Systematically handle all components of an Input object.
+        new_input = Input()
+
+        # > First simple keywords will be handled by merging the two lists of simple keywords, dropping duplicates
+        new_keyword_list = list(
+            set(
```
haneug: I dislike that for the simple keywords we will get any order. It makes the resulting input files not nice to look at for the user. You could use `list(dict.fromkeys(...))` and use an ordered dict to preserve the order. Of course it would still be input1 + input2 but I think that would still look nicer than any order.

haneug (approved):
Nice addition, please consider my suggestion if it is an improvement or not!

haneug:
Please also add the PR # to the CHANGELOG and to the PR itself for consistency. And delete the section about Keep a CHANGELOG.

PR #290, opened 2026-09-03
feat: add `Output.get_final_energy_components()`

### LINKED ISSUE #281: Missing components in `Output.get_energies()`

haneug:
## Closes Issues
Closes #281

- [ ] Write proper CHANGELOG
- [ ] Split enum into two enums

## Description
- Adds a new getter method `Output.get_final_energy_components()` which combines the energies of `Output.get_energies` and additional separate corrections (dispersion and gcp)
- Also adds an enum for the dict entries in both energy methods. It is not meant to be used as the actual API but as a reference for the available energies.

---

### Added 

- Added `Output.get_final_energy_components()` which resolves the final energy contributions. (#XXX)


nakul680 (approved):
Looks good!

PR #293, opened 2026-09-07
fix: Make OPI runner thread safe by copying the env for each run()

### LINKED ISSUE #292: OPI Runner not thread safe
**Describe the bug**
The OPI runner is currently not thread safe and when running multiple jobs in parallel ORCA will not be found since a global env is used and this is cleared after a job is finished.

haneug:
## Closes Issues
Closes #292

## Description
- Make the OPI runner thread safe by copying the env for every run with a `BaseRunner._build_env()` function. The need for the `_orca_environment` decorator is not given anymore since we do not need a clean up. Using a decorator would also force us to expose the `env` argument in the `BaseRunner.run()` function which is another downside in my opinion.

---

### Fixed 

- Made OPI runner thread safe by copying the env for each run (#293).


timmyte (approved):
Looks good! Thanks for adding that.
But please include release notes. It's worth mentioning the new variant is thread-safe.

